PR #3, opened 2026-07-08
[Fix] Treat a clarification question as a valid end of turn in Slack hooks

daniel-lxs:
## Problem

When an agent asks the user a question via `send_chat_reply` purpose `clarification` and ends its turn, the stop hook blocks (`current_turn_nonterminal_reply` — only `closeout` counted as terminal). The harness then injects up to **3** closeout reminders, and the model posts redundant echo messages right after its own question: *"Paused here for now — I'm waiting on your answer to the question above."*

Investigation notes (from the OpenRoomote debugging session; ported here per repo targeting):
- This is the only live echo path: the `request_user_input` branches in both hooks are dead code (nothing writes that tool into the satisfaction state), and the structured question tool blocks the turn in-flight so the stop hook never runs while it waits.
- A second echo source was caught live during e2e verification: after a closeout, a `todowrite` bookkeeping call staled the terminal reply (the exemption list only knew the legacy `update_plan` name), forcing a redundant "closed this turn" follow-up.

## Fix

Principle: the hooks exist to prevent turns from ending **silently**. A clarification question is a visible handoff exactly like a closeout — the turn ends waiting on the user's answer — so it now counts as terminal, with the same staleness rule as closeout (non-Slack work after the question invalidates it). `ack` and `progress` still block; automation no-turn closeout enforcement is untouched.

- `recordChatReplySatisfaction` stamps terminal state for `clarification` replies
- Stop hook accepts clarification in terminal-satisfaction checks (incl. legacy states written by pre-deploy workers)
- Silence hook mirrors the same policy; OpenCode `todowrite`/`todoread` count as closeout bookkeeping and no longer stale a terminal reply
- `send_chat_reply` description updated: a clarification ends the turn when the next step depends on the user's answer — don't follow it with a separate "waiting on your answer" message

## Verification

- Hook-script tests exercise the real generated scripts: clarification at turn end → allow (`terminal_reply_satisfied`); non-Slack work after the question → block (`current_turn_terminal_reply_stale`); ack/progress still block; todowrite/todoread don't stale a closeout
- Verified end-to-end on a live planner task via the mock Slack harness: question turn ends cleanly with the terminal state stamped and no echo for 3+ minutes (previously within one minute), and the answer round-trip completes with a single closeout
- `apps/worker` run-task + satisfaction tests: 228 passed; `tsc --noEmit` and prettier clean


daniel-lxs:

PR #9, opened 2026-07-08
[Improve] Bring the Azure DevOps source-control settings section to parity

mrubens:
## Summary

Brings the Azure DevOps section of Settings > Source Control to parity with GitLab/Gitea, and fixes two UI bugs surfaced while testing it.

- **Save-time PAT validation**: newly saved `ADO_TOKEN` values are now validated against the organization `connectionData` API before persisting, matching GitLab/Gitea. Definitive rejections block the save — including Azure DevOps' quirk of answering rejected PATs with a 203 sign-in page instead of a 401. Connectivity failures don't block, and validation is skipped when no organization is available yet.
- **Setup instructions no longer hidden by unrelated config**: the settings page treated a provider as "configured" when *any* field was satisfied. `ADO_TENANT_ID` accepts `ROOMOTE_AUTH_MICROSOFT_TENANT_ID` as a fallback, so deployments using Microsoft sign-in showed a bare credentials form with no setup instructions. Providers now count as configured only when all required fields are satisfied.
- **Expanded sections always render their credentials form**: the page tracked a single "configuring" provider, so opening one provider's setup dropped the form out of any other open section (instructions with nowhere to paste credentials). The form now always renders in expanded sections and the extra state is gone.
- **Copy fixes**: correct article ("Create an Azure DevOps personal access token"), plus organization-slug, PAT-scope, and bot-account guidance in the creation hint.

Also updates `.agent-guidance/features/ado-connection-setup.md` to document the new validation and gating behavior.

## Testing

- New unit tests: `validateAdoToken` (valid / 203-rejection / network-unknown), ADO config-validation wiring (invalid token blocks save, saved-org fallback, skip without org), and component regressions for both UI bugs
- Verified live on a dev deployment: ADO shows the full setup walkthrough, and Gitea + ADO setup forms stay visible when both are open
- `pnpm lint:fast`, `pnpm check-types:fast`, and `pnpm knip` all pass

### DIFF
--- a/.agent-guidance/features/ado-connection-setup.md
+++ b/.agent-guidance/features/ado-connection-setup.md
@@ -1,7 +1,7 @@
 ---
 title: Azure DevOps Connection Setup
 status: active
-last_reviewed: 2026-07-05
+last_reviewed: 2026-07-08
 owner: engineering
 summary: Operator-facing setup guide for connecting Azure DevOps to Roomote through deployment-token-backed source control, service hooks, and optional Entra-backed commenter account linking.
 ---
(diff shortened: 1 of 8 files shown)


mrubens:
Addressed the review suggestion in ab2010a5: `isProviderConfigured` now reads the server-computed `providerStatus.configSatisfied` instead of re-deriving the required-fields check client-side, and the component test mocks were simplified to match.

PR #16, opened 2026-07-09
[Improve] Correct Railway Composer notes after configuring the live prompts

mrubens:
Follow-up to #10, aligning the spec and operator guide with the published templates (main `Rj2cFo`, develop `bP3Lsu`).

**Composer corrections** (from configuring the `ROOMOTE_APP_URL` deploy-screen prompt on both templates and verifying on the live deploy screens):
- Railway **does** have a template Duplicate action; the maintainer note no longer claims otherwise.
- The prompt surfaces inside the api service's **Configure** step, and the verification checklist now says where to look.

**Spec adoption of two variables the published templates already ship:**
- `SETUP_TOKEN=${{secret(32)}}` on api (referenced by the other app services). This is not optional anymore: `apps/web/src/lib/server/setup-token.ts` refuses tokenless first-admin bootstrap outside local development, so a template without it cannot complete setup at all. The spec previously said it was intentionally unset — that note was stale against the app.
- `ROOMOTE_PING_BASE_URL=https://ping.openmote.dev` — anonymous-analytics and version-check pings (image default is `ping.roomote.dev`); analytics opt-out stays available in the setup wizard and Settings → Misc.

### DIFF
--- a/deploy/railway/README.md
+++ b/deploy/railway/README.md
@@ -151,6 +151,8 @@ S3_ACCESS_KEY_ID=roomote
 S3_REGION=us-east-1
 S3_BUCKET_ARTIFACTS=roomote-artifacts
 S3_AUTO_CREATE_BUCKET=true
+SETUP_TOKEN=${{secret(32)}}
+ROOMOTE_PING_BASE_URL=https://ping.openmote.dev
 DEFAULT_COMPUTE_PROVIDER=modal
 EXCLUDED_COMPUTE_PROVIDERS=docker
 DATABASE_URL=${{Postgres.DATABASE_URL}}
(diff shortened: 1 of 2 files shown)


mrubens:
Live-template state as of this PR: both published templates (main `Rj2cFo`, develop `bP3Lsu`) now match this spec exactly — the `ROOMOTE_APP_URL` deploy prompt is configured, `SETUP_TOKEN`/`ROOMOTE_PING_BASE_URL` are present, and the web/controller/bullmq shared blocks use `${{api.*}}` references instead of repeated raw values. A scratch deploy verified the two-hop reference resolution produces real URLs in the app services' variables (checklist run 2026-07-08).

PR #20, opened 2026-07-09
[Improve] Ship the preview-proxy service in the Railway templates

mrubens:
Moves `preview-proxy` from "optional service you add by hand" to a standard service in both Railway templates.

## Why it's safe as a default

[`apps/preview-proxy/src/config.ts`](apps/preview-proxy/src/config.ts) only hard-requires `ROOMOTE_APP_URL`, `DATABASE_URL`, and `REDIS_URL` — all already in the shared env block — and the `PREVIEW_*` variables are `emptyStringDefault()` in `@roomote/env`. So the service boots healthy with previews unconfigured and sits inert until DNS points at it. Cost to non-preview deployers is one idle Node process.

## What changes

- `template.yaml`: `preview-proxy` becomes a standard service using the shared env block verbatim. The three `PREVIEW_*` variables move to **api** as empty values (annotated to be marked **optional** in the Composer so the deploy screen doesn't prompt for them), with `${{api.*}}` references in the shared block — same anchor pattern as everything else.
- README: topology row un-optionaled; "Enabling live previews" drops the add-the-service step and becomes config-only (attach the two custom domains, fill three values on api, opt in at Settings → Live Previews).

## Follow-ups after merge

- Mirror into both published templates via the Composer (add the service, add the three optional empty variables on api, paste the updated shared block into all four non-api app services) and re-run the scratch-deploy check — including that preview-proxy reports healthy with the variables empty and that the deploy screen does not prompt for them.
- The onboarding script (#19) keeps working against both old and new deployments — its reference upserts become no-ops on template-created services, and its "preview-proxy missing" guard now only triggers for pre-change deployments.


mrubens:
Addressed the review checklist: the pin/upgrade instructions in deploy/railway/README.md and the design note in deploy/railway/template.yaml now count five app-service image fields (preview-proxy included). Also merged develop to pick up the notify-ops workflow changes.

mrubens:
Addressed both follow-ups: the Costs control-plane list and the canonical-origin variable note now include preview-proxy.

PR #30, opened 2026-07-09
[Fix] Unblock Azure DevOps onboarding: environment config, cloning, and webhooks

daniel-lxs:
## Summary

Four fixes found by exercising a fresh Azure DevOps connection end to end (org + PAT → repo sync → environment mapping → service hooks → manual task → PR automation → comment mentions). Each fix was verified live against a real `dev.azure.com` organization.

### 1. Environment configs rejected Azure DevOps repositories
`environmentRepositoryConfigSchema` required exactly `owner/repo`, but ADO full names are always `organization/project/repo` — so a synced ADO repo could never be added to an environment through YAML create/update, which also gates PR automation. GitLab subgroup repos (3+ segments) had the same latent problem. The regex now accepts two or more non-empty slash-separated segments; everything downstream already keys off exact `fullName` equality.

### 2. Worker clones failed with `could not read Password`
ADO `remoteUrl` embeds the organization as URL userinfo (`https://org@dev.azure.com/...`). Git's `insteadOf` is a literal prefix match, so the worker's credential-proxy rewrite never fired and git prompted for a password it couldn't get. Fixed at both ends with a shared `stripCloneUrlUserInfo` helper: sync stores a clean `cloneUrl`, and the worker strips userinfo defensively before cloning (covers rows synced before this fix).

### 3. Every ADO webhook delivery 500'd at the web proxy
Azure DevOps service hooks send `Expect: 100-continue`; undici's `fetch` rejects the header outright (`UND_ERR_NOT_SUPPORTED`), so the `/api/webhooks/[...path]` forward threw before reaching the API. The `Expect` header is now dropped along with hop-by-hop headers — the 100-continue handshake is meaningless to replay on a proxied request.

### 4. ADO comment webhooks failed schema validation
Real `ms.vss-code.git-pullrequest-comment-event` deliveries carry the comment object directly as `resource` — even with `resourceVersion: 1.0` pinned on the subscription — not the documented `resource: { comment, pullRequest }` nesting (confirmed against the raw delivery body in ADO's notification history). A new normalization step detects the flat shape, extracts repository + PR id from `resource._links`, rehydrates the pull request via a new `getAdoPullRequest` helper, and hands the documented shape to the existing schema and handler unchanged.

## Validation

- New unit tests for all four fixes (`command-schema`, `source-control`, `ado api`, webhook proxy route, `normalizeCommentWebhook`)
- `pnpm lint:fast`, `pnpm check-types:fast`, `pnpm knip` all pass
- Verified live end to end: repo sync stores clean clone URLs, worker clones through the credential proxy, `pullrequest.created`/`updated` deliveries return 200 and enqueue review tasks, review comments post back to the ADO PR thread, `@roomote` comment mentions parse and route correctly, and PR completion status updates deliver cleanly


daniel-lxs:

PR #89, opened 2026-07-10
[Fix] targetBranch is no longer required when updating an existing pull request

daniel-lxs:
## Problem

The roomote MCP tool `manage_source_control` with `action: create_or_update_pull_request` required `targetBranch` even when the PR already exists. Models repeatedly omit it on update calls and burn 1-2 failed tool calls per round on `targetBranch is required for create_or_update_pull_request` (observed in staging task 2o885wsfv775m and prod task transcripts on 2026-07-10).

## Fix

`targetBranch` is now optional when an open PR already exists for the source branch:

- **SDK mutation handler** (`packages/sdk/src/server/lib/pull-requests/source-control-pull-requests.ts`): the input schema makes `targetBranch` optional. When it is omitted, every provider (GitHub, GitLab, Gitea, ADO) looks up the open PR/MR by source branch alone and defaults to that PR's current base ref; the base is never changed on update unless `targetBranch` is passed explicitly (explicit retargeting on GitHub still works).
- **Creation still requires it**: creating a new PR without `targetBranch` throws a 400 whose message says exactly when the field is required ("targetBranch is required to create a pull request: no open pull request was found for source branch ... Retry with targetBranch set ...").
- **Ambiguity is refused, not guessed**: an omitted `targetBranch` with multiple open PRs from the same source branch (different bases) returns a 409 listing the candidate target branches instead of silently updating an arbitrary one.
- **Worker tool** (`apps/worker/src/mcp/roomote-mcp-server/source-control.ts`): the eager `targetBranch` validation is removed; blank values are normalized to absent (models emit empty strings for unused optional fields). The tool description and the `targetBranch` field description now state when the field is required.
- The mutation result now includes the effective `targetBranch` (input value or inherited base), which also feeds `prBaseRef` persistence on `task_pull_requests`.

## Tests

- SDK: update-without-`targetBranch` inherits the existing base for all four providers (and persists it as `prBaseRef`); create-without-`targetBranch` rejects with the actionable 400; multiple open PRs for the branch reject with 409; explicit `targetBranch` still retargets on GitHub.
- Worker: new `source-control.test.ts` covering omitted/blank/explicit `targetBranch` forwarding, the remaining required fields, and platform error surfacing.
- `pnpm test` green for `@roomote/sdk` (409), `@roomote/worker` (1261), `@roomote/api` (773); full `pnpm check-types` and `pnpm lint` pass. `pnpm knip` failures are pre-existing and untouched by this change.

### DIFF
--- a/apps/worker/src/mcp/roomote-mcp-server/__tests__/source-control.test.ts
+++ b/apps/worker/src/mcp/roomote-mcp-server/__tests__/source-control.test.ts
@@ -1,0 +1,212 @@
+import { handleManageSourceControl } from '../source-control.js';
+import * as tasksApiClient from '../tasks-api-client.js';
+import type {
+  RoomoteConfig,
+  SourceControlPullRequestResponse,
+} from '../types.js';
+
+vi.mock('../tasks-api-client.js');
+
+const config: RoomoteConfig = {
+  token: 'test-token',
+  platformApiUrl: 'https://test-api.example.com',
+};
+
+function pullRequestResponse(
+  overrides: Partial<SourceControlPullRequestResponse> = {},
+): SourceControlPullRequestResponse {
+  return {
+    success: true,
+    action: 'updated',
+    provider: 'github',
+    repositoryFullName: 'acme/web',
+    number: 12,
+    url: 'https://github.com/acme/web/pull/12',
+    title: '[Feature] X',
+    targetBranch: 'develop',
+    draft: false,
+    warnings: [],
+    ...overrides,
+  };
+}
+
+describe('handleManageSourceControl create_or_update_pull_request', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('forwards an update without targetBranch so the platform defaults to the existing PR base', async () => {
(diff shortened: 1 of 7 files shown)


daniel-lxs:
Addressed both review findings in dc584b9:

1. **Gitea pagination** — `listGiteaPullRequests` now walks pages (`page`/`limit=50`) instead of trusting the first page, since Gitea's pulls listing has no head/base filter. It early-exits once it has what it needs: one match when an explicit `targetBranch` pins the unique head+base pair, two when an omitted `targetBranch` only needs enough to prove ambiguity, with a 40-page runaway guard. Tests cover finding the PR on page 2 with a full non-matching first page, and stopping after page 1 when a full page already contains the explicit-target match.

2. **GitHub explicit-target path** — the conditional `base` on `pulls.update` was dead code: the base-scoped lookup can only return a PR that already targets the requested base, so it's removed and the update never sends `base`. The impossible retarget test is replaced with two tests pinning the real semantics: an explicit `targetBranch` updates the matching PR without touching its base, and opens a new PR against that base when the branch's open PR targets a different one (GitHub allows same head to different bases).

PR #94, opened 2026-07-10
[Fix] Recover subagent spawns that never settle after the child finishes

daniel-lxs:
Replaces the removed time-based subagent watchdogs (#90, #92) with a detector that cannot false-positive on thinking models, because it keys on a **state contradiction instead of a clock**:

- OpenCode reports a tracked foreground child session **terminal** (`session.idle` / `session.error`) — normally the spawn's task tool part settles within seconds from the child's completion.
- If the part is still unsettled after a grace window (default **2 min**, `ROOMOTE_SUBAGENT_SETTLEMENT_GRACE_MS`), the spawn has leaked inside OpenCode (the class behind task `2vz6czf1c0n6x`'s abandoned subagents) and the parent turn would wait on it forever.
- Recovery aborts **that one finished child session by its known id** — no list-all-children fallback, so no collateral on siblings. The abort drives OpenCode's own task-tool cancellation path, settling the spawn so the parent resumes with a visibly failed spawn it can re-run.

Safety properties:
- A **thinking child can never trip it** — a busy session isn't idle; silence is never a signal.
- **Any further child event cancels** the pending recovery (provider retries that pick the session back up are safe).
- **Background launches unchanged** (idle = done). Turn teardown / settlement / dispose clear the timer.
- Recovery acts on the child only; it never completes or aborts the parent turn.

## Tests
7 new settlement-recovery tests (leak → single targeted abort; settle-within-grace, child-revival, background, turn-finish, and dispose all cancel; child error arms the same path) alongside the no-time-based-abort invariants. Full harness suite: **144 tests pass**; typecheck / eslint / prettier clean.

### DIFF
--- a/apps/worker/src/sandbox-server/lib/harnesses/__tests__/opencode-server-subagent-watchdog.test.ts
+++ b/apps/worker/src/sandbox-server/lib/harnesses/__tests__/opencode-server-subagent-watchdog.test.ts
@@ -574,6 +574,92 @@ describe('OpenCode subagent settlement recovery', () => {
     }
   });
 
+  it('never aborts a child whose latest assistant message is still in flight', async () => {
+    const { client, harness, logger } = createHarness();
+
+    try {
+      await connectHarness(harness, client);
+      vi.useFakeTimers();
+      await armSpawn(client, harness);
+
+      // The child looks terminal, but its persisted state says it is still
+      // mid-message — a silent revival or a lookup we cannot trust. Recovery
+      // must keep waiting instead of killing possibly-live work.
+      client.messages.mockResolvedValue([
+        {
+          info: {
+            id: 'msg_child_live',
+            sessionID: 'ses_child_1',
+            role: 'assistant',
+            time: { created: 1 },
+          },
+          parts: [],
+        },
+      ] as unknown as OpenCodeSessionMessage[]);
+
+      await client.emit({
+        type: 'session.idle',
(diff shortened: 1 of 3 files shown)


daniel-lxs:
Hardened against false positives per Dan's requirement (healthy subagents must never be killed):

**Residual false-positive vectors found by adversarial review, now closed:**
1. *Silently revived child* — a child that idles, then gets picked back up (task-tool resume, provider retry) is normally protected by any-event-cancels; if the busy transition emitted no event and the child then thought silently past the grace, the old code would abort a live session.
2. *Grace asymmetry* — a false abort kills real work; a slow leak recovery only delays an already-stuck spawn.

**Changes:**
- Grace default 2 min → **10 min** (`ROOMOTE_SUBAGENT_SETTLEMENT_GRACE_MS`).
- **Pre-abort verification:** at expiry, fetch the child's messages and abort only if its **latest assistant message is completed** — finished work + unsettled spawn = provable leak. An in-flight message, or a failed lookup, never aborts: recovery logs and re-checks a grace window later, indefinitely. A settle that races the verification also cancels the abort.

Net bias: the only way a child can be aborted is opencode reporting it terminal **and** its persisted state showing finished work **and** the spawn still unsettled 10+ minutes later **and** zero events from the child in between. Thinking models fail the first condition (busy ≠ idle); revived children fail the second or emit events. 147 harness tests pass, including the two new false-positive guards (in-flight message → no abort then recovery after completion; verification failure → no abort).

PR #95, opened 2026-07-10
[Fix] Derive SLACK_AUTH_URI from ROOMOTE_APP_URL when unset

mrubens:
## Problem

When `SLACK_AUTH_URI` is unset, the Slack account-linking onboarding DM silently fails. The env schema defaults it to an empty string, so the "Link accounts" button URL becomes `?state=<token>` — Slack rejects the entire `chat.postMessage` with `invalid_blocks` ("invalid url"), and unlinked users who DM the bot get no response at all.

This hit the roomote-nightly Railway deployment on 2026-07-10: none of the fleet deployments set `SLACK_AUTH_URI` (the ops provision script only sets `ROOMOTE_APP_URL`).

## Fix

When `SLACK_AUTH_URI` is unset or empty, derive it from `ROOMOTE_APP_URL` as `${ROOMOTE_APP_URL}/api/slack/auth` — the route the web app already serves for completing the link (`apps/web/src/app/api/slack/auth/route.ts`). An explicitly set `SLACK_AUTH_URI` still wins.

The fallback lives in `buildRoomoteRuntimeEnv` (packages/env) rather than at the button-building site, so every service that constructs env — including `rehydrateEnv` and the web bootstrap — gets it, matching the existing derivation pattern for `SLACK_API_TIMEOUT_MS` and `DOCKER_WORKER_IMAGE`.

## Changes

- `packages/env/src/index.ts` — derive `SLACK_AUTH_URI` from `ROOMOTE_APP_URL` when unset/empty (trailing slashes trimmed)
- `packages/env/src/__tests__/index.test.ts` — tests for the fallback (unset, empty, trailing slash) and explicit-override precedence
- `apps/docs/environment-variables.mdx` — document the default

## Validation

- `pnpm --filter @roomote/env exec vitest run src/__tests__/index.test.ts` — 53 passed
- `pnpm --filter @roomote/slack exec vitest run src/__tests__/block-kit.test.ts` — 21 passed
- `pnpm lint` and `pnpm check-types` — clean across all workspaces

No fleet config change needed: deployments already set `ROOMOTE_APP_URL`, so the next deploy with this commit fixes the silent DM failure.

### DIFF
--- a/apps/docs/environment-variables.mdx
+++ b/apps/docs/environment-variables.mdx
@@ -242,7 +242,7 @@ as per-task auth tokens or workspace paths.
 | `SLACK_CLIENT_ID`                      | Slack app/auth    | Slack OAuth client ID. Also accepted for Slack sign-in setup.                   |
 | `SLACK_CLIENT_SECRET`                  | Slack app/auth    | Slack OAuth client secret. Also accepted for Slack sign-in setup.               |
 | `SLACK_REDIRECT_URI`                   | Optional          | Slack redirect URI override.                                                    |
-| `SLACK_AUTH_URI`                       | Optional          | Slack auth URI override.                                                        |
+| `SLACK_AUTH_URI`                       | Optional          | Slack auth URI override. Defaults to `ROOMOTE_APP_URL` + `/api/slack/auth`.     |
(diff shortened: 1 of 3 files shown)


mrubens:
@openmote should we do the same for SLACK_REDIRECT_URI?

PR #98, opened 2026-07-10
[Fix] Subagent rows disappear from the transcript on refresh

daniel-lxs:
Fixes the display bug Matt and Dan hit today: a running subagent shows while live events stream, then **vanishes on refresh** "until an update comes in".

## Root cause
Spawn-row visibility keyed on the live-only `subagentActivity` payload. That field is emitted via `toolUpdate` → `output(...)` (live stream) and **never persisted**, so a transcript rebuild after refresh drops it — and both visibility gates (`showInternalMessages` gate and the narration pending-hide) then hid the row until the next live activity event arrived (≤5s for an active child, forever for one that settled or was abandoned).

## Change
Visibility now keys on the stable payload shape: task-tool spawn rows (`kind: 'subagent'`, no bound `receiverThreadIds`) always render inline — default and narration modes, with or without live activity. Rendering already handled activity-less rows (compact title row), so only the visibility rules changed:
- `isSubagentSpawnRowMessage` replaces the `hasSubagentLiveActivity` exemption in the internal-messages gate.
- The narration `shouldHidePendingSubagentMessage` gate is removed (helpers deleted).
- Thread-bound subagent messages (`receiverThreadIds` populated) keep their existing behavior — they surface via the ActiveSubtasksList, not inline rows.

## Tests
New `subagent-visibility` suite: rebuilt (activity-less) spawn rows render in both modes; thread-bound and non-spawn subagent payloads stay hidden without debug UI; debug UI still reveals everything. Two existing tests that pinned the old hide behavior updated to pin the new one. Task-page suite: **505 tests pass**; typecheck / eslint / prettier clean.

### DIFF
--- a/apps/web/src/app/(sandbox)/task/[taskId]/messages/acp/__tests__/subagent-visibility.client.test.ts
+++ b/apps/web/src/app/(sandbox)/task/[taskId]/messages/acp/__tests__/subagent-visibility.client.test.ts
@@ -1,0 +1,98 @@
+import type { AcpToolCallUiMessage, AcpUiMessage } from '../types';
+import { buildAcpRenderBlocks } from '../render-blocks';
+
+// A spawn row as it comes back from a transcript rebuild: persisted envelope
+// data only, no live-only `subagentActivity` payload.
+function rebuiltSpawnMessage(params: {
+  id: string;
+  status?: 'completed' | 'failed' | 'in_progress';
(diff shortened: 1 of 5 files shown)


daniel-lxs:
Added a second commit fixing the `generate-summary` test expectation that #87 left stale — develop CI has been red since it merged, which is what failed this PR's first CI run.

PR #104, opened 2026-07-10
Finish CloudJob to TaskRun migration

mrubens:
## Summary
- replace CloudJob/CloudTask compatibility types and IDs with TaskRun/run vocabulary
- move REST, SDK, tRPC, web, queue, controller, worker, and integration surfaces to task-run naming
- rename runtime metrics, logs, auth tokens, events, and files without public compatibility aliases
- merge the latest origin/develop schema and web updates
- relax one Slack routing test timeout that repeatedly exceeded 5 seconds only under full-suite load

## Validation
- pnpm lint
- pnpm check-types
- pnpm test (24/24 packages)
- pnpm knip
- legacy vocabulary scan
- git diff --check


mrubens:
Addressed both deploy-time compatibility notes in 6a2a3c35:

- The task-run queue now drains `queue:cloud-jobs` before the current queue and deduplicates scopes across both keys. New work is still written only to `queue:task-runs`, so the old list drains naturally without preserving a public alias.
- Run-token validation privately accepts signed pre-deploy `t: cj` tokens and immediately normalizes them to `tokenType: run`. The exported payload schema and token minting remain strictly `t: run`, so this only covers in-flight tokens until their existing TTL expires.

Added focused queue and token rollover tests. Full `pnpm check` passes.

PR #105, opened 2026-07-10
Harden application container privileges and secret scope

mrubens:
## Summary

- split the control plane into service-specific non-root runtime targets and publish each target independently
- replace shared `.env` injection with explicit web, API, controller, BullMQ, preview-proxy, and migration environment contracts
- run application services with read-only filesystems, tmpfs scratch space, dropped capabilities, and `no-new-privileges`
- move raw Docker socket access into a digest-pinned proxy on an internal controller-only network
- keep the Docker CLI and worker-release assets controller-only; remove Git, GitHub CLI, Python, and OpenCode from application runtimes
- add a CI-enforced production Compose security contract and update self-hosting/Coolify documentation

## Security notes

The socket proxy denies unrelated Docker API sections and removes the raw socket from the controller. Container lifecycle access remains host-critical because the controller must create task workers; the proxy and Docker host should still be treated as trusted infrastructure. A purpose-built controller daemon would be a stronger future boundary.

## Validation

- `deploy/scripts/validate-compose-security.sh`
- `pnpm --filter @roomote/env exec vitest run src/__tests__/index.test.ts` (56 tests)
- `pnpm --filter @roomote/env check-types`
- `pnpm lint:fast`
- `pnpm check-types:fast`
- `pnpm knip`
- Dockerfile build checks for app and worker images
- built and inspected controller and migration runtime images
- smoke-tested proxy-backed Docker run/cp/exec/inspect/rm flow and verified the volumes API returns HTTP 403


mrubens:
Addressed in ff644f4a.

The compatibility target now copies the Next.js standalone output, static assets, and public tree with `--chown=roomote-app:roomote-app`. This fixes the ownership at the image layer for Coolify, Fly, Railway, and any other platform using `roomote-app`, rather than adding a Coolify-only mount.

Verified by building `runtime-app`, running it as its configured non-root UID/GID 10000, and successfully creating/writing `/roomote/apps/web/.next/cache/review-test/write-check`. Dockerfile build checks and the pre-push lint/typecheck/knip gates also pass.

PR #106, opened 2026-07-10
[Chore] Validate deployment artifacts in CI

mrubens:
## What changed

- build both app and worker images in pull-request CI
- validate every maintained Compose shape plus Railway, Render, Coolify, and Fly against a shared deployment catalog
- add production controller and BullMQ health checks
- gate image publishing on a production deployment lifecycle: generated secrets, migrations, setup/health probes, Docker-backed task launch, prior-channel upgrade, rollback, backup, and fresh-volume restore
- retain native amd64 and arm64 release builds
- pin critical infrastructure images and external GitHub Actions to immutable digests/commit SHAs

## Why

Deployment artifacts were not first-class CI inputs, so `docker compose up --wait` could succeed while the execution plane was unhealthy and release-only deployment paths could drift unnoticed.

## Validation

- `pnpm deployment:validate`
- `pnpm --filter @roomote/cloud-agents check-types`
- `pnpm --filter @roomote/cloud-agents lint`
- `pnpm knip`
- pre-push `pnpm lint:fast`, `pnpm check-types:fast`, and `pnpm knip`
- production app and worker image builds on linux/arm64
- prior `develop` image → candidate migration → rollback → candidate
- Docker-backed task launch through the production controller
- backup restored onto fresh Compose volumes with all services healthy

## Compatibility note

`v0.0.1` predates the current rewritten Drizzle migration baseline and cannot be used as a direct migration predecessor. For the first gated tagged release, CI tests the prior published `main` candidate. Subsequent tagged releases use the immediately preceding immutable release tag.

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -26,19 +26,19 @@ runs:
   using: 'composite'
   steps:
     - name: Setup pnpm
-      uses: pnpm/action-setup@v4
+      uses: pnpm/action-setup@f40ffcd9367d9f12939873eb1018b921a783ffaa # v4
       with:
         version: ${{ inputs.pnpm-version }}
 
     - name: Setup Node.js
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
       with:
         node-version: ${{ inputs.node-version }}
         cache: pnpm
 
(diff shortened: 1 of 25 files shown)


mrubens:
Addressed the review nit in 30a9a953: the Fly process validation now uses one shared assertion for every runtime service. Re-ran `pnpm deployment:validate`; pre-push lint, fast type checks, and Knip also pass.

PR #109, opened 2026-07-10
Harden Docker task sandbox isolation and resource limits

mrubens:
## Summary

- enforce configurable CPU, memory/swap, PID, writable-layer, and bounded log limits for Docker workers
- create a dedicated network per task and attach only the API and preview proxy as trusted peers
- enforce public-only or no-egress policies with private, cloud-metadata, sibling-network, and direct gateway protection
- reap stale provisioning containers and task networks after controller restarts
- document the new defaults and self-host configuration across Compose and Coolify deployments

## Testing

- `RUN_DOCKER_SANDBOX_INTEGRATION_TESTS=true pnpm --filter @roomote/controller exec vitest run src/compute-providers/__tests__/docker-sandbox-isolation.integration.test.ts`
- `pnpm exec dotenvx run -f .env.test -- pnpm --filter @roomote/controller exec vitest run`
- `pnpm exec dotenvx run -f .env.test -- pnpm --filter @roomote/env exec vitest run`
- `pnpm lint:fast`
- `pnpm check-types:fast`
- `pnpm knip`

The Docker integration test verifies enforced HostConfig limits, trusted-service reachability, sibling sandbox/preview/credential denial, immutable egress routes, and crash-recovery cleanup.


mrubens:
Addressed all four review findings in `20ac9b03`:

- Preview proxy is optional again: only the API is required, while `preview-proxy` is attached when present.
- Writable-layer limits now retry without `--storage-opt size` only when Docker reports that the storage driver cannot enforce the quota; other startup failures still fail closed. The fallback emits an operator warning, and the docs explain how to retain a hard host-level disk bound.
- Local-development egress policy continues blocking metadata ranges but no longer blackholes RFC1918 routes, preserving `host.docker.internal` on Docker Desktop. Self-hosted networked deployments still block private ranges and the bridge gateway.
- Worker process-list matching is consolidated into one shared helper used by both startup ownership checks and stale-resource cleanup.

Validation:
- controller suite: 82 passed, 2 skipped
- real-Docker isolation suite: 2 passed, including sibling/credential isolation, trusted API/preview reachability, crash cleanup, and live `host.docker.internal` connectivity
- `pnpm lint:fast`
- `pnpm check-types:fast`
- `pnpm knip`
- no leftover test containers or networks

mrubens:
Addressed the latest review findings in `a705dfa9`:

- Docker disk quotas now fail closed by default. `DOCKER_WORKER_ALLOW_UNBOUNDED_DISK=true` is an explicit operator opt-in intended only when an equivalent host-level quota exists.
- The controller now reconciles replacement API and preview-proxy containers onto every running task network during startup and periodic cleanup. The real-Docker suite replaces the API container and verifies existing workers reconnect without gaining sibling access.
- Docker inspect/exec failures are no longer collapsed into “resource missing” during reaping, so transient daemon or socket-proxy failures cannot trigger destructive cleanup.

Validation:
- full controller suite: 88 passed, 2 skipped
- environment suite: 57 passed
- real-Docker isolation suite: 2 passed
- production Compose security contract
- `pnpm lint:fast`
- `pnpm check-types:fast`
- `pnpm knip`
- no leftover test containers or networks

PR #129, opened 2026-07-10
[Fix] OpenCode SDK server fails to start on Railway: root-owned HOME artifacts in the inference image

mrubens:
## Problem

On nightly (Railway), the LLM router falls back on every request, and task title refresh / work-kind classification fail:

```
[LLM Router] Routing fallback ... reason="OpenCode SDK server exited before it was ready ... code 1"
PermissionDenied: FileSystem.open (/tmp/.local/share/opencode/log/opencode.log)
```

Root cause: `runtime-inference-base` (#114) installs OpenCode and runs `opencode --version` **as root** with the image's `HOME=/tmp`. That bakes root-owned `/tmp/.npm` and `/tmp/.local/share/opencode/` into the image layer. The final stage runs as the hardened non-root `roomote-app` user (#105) with the same `HOME`, which cannot write inside those root-owned directories, so the in-process OpenCode SDK server dies at startup.

Why nothing caught it:
- **Compose deployments** (and the deployment acceptance smoke test) mount a fresh tmpfs over `/tmp`, masking the baked artifacts entirely — the skew only bites hosts that run the raw image, i.e. Railway.
- The smoke test's tool check was `command -v opencode`, which locates the binary without executing it.
- #114 was never deployed to Railway until the publish pipeline unblocked today (#125), so the first production exposure was `develop-f68732c7` at 20:10 UTC.

## Fix

- `.docker/app/Dockerfile`: remove `/tmp/.npm`, `/tmp/.local`, and `/tmp/.cache` in the same `RUN` layer that installs and version-checks OpenCode, so the image ships a pristine `/tmp` and the runtime user creates its own `HOME` state.
- `deploy/ci/deployment-smoke.sh`: the candidate check now executes `opencode --version` as the container user instead of just locating it. This verifies runtime executability (startup writes state under `HOME`); the comment documents that the tmpfs masks the raw-image case, which the Dockerfile layer now prevents.

## Testing

- `bash -n` on the smoke script. Local Docker daemon is unhealthy on this machine, so the image build is validated by this PR's Docker Build checks and, post-merge, by the (re-armed) deployment acceptance gate on the develop publish run.
- Verification on nightly after deploy: api logs should show no `PermissionDenied ... opencode.log`, and `[LLM Router]` lines should stop reporting `phase="fallback"` with the SDK-server-exit reason.

### DIFF
--- a/.docker/app/Dockerfile
+++ b/.docker/app/Dockerfile
@@ -282,19 +282,21 @@ FROM runtime-base AS runtime-inference-base
 
 ARG OPENCODE_CLI_VERSION=1.17.8
 
-# The install and version check run as root with the image's HOME=/tmp, which
-# bakes root-owned /tmp/.npm and /tmp/.local (OpenCode's data dir) into the
-# layer. The runtime user shares that HOME and must be able to create those
-# paths itself — on hosts that run the raw image without a tmpfs over /tmp
-# (e.g. Railway), leftover root-owned dirs make OpenCode's log open fail with
-# PermissionDenied and the SDK server exits before ready. Strip them in the
-# same layer.
-RUN npm install --prefix /opt/opencode-cli --no-save --no-package-lock \
+# The install and version check run as root; with the image's HOME=/tmp they
+# would bake root-owned dotdirs (npm cache, OpenCode's data/config/cache
+# dirs) into the layer. The runtime user shares that HOME and must be able to
+# create those paths itself — on hosts that run the raw image without a tmpfs
+# over /tmp (e.g. Railway), leftover root-owned dirs make OpenCode's log open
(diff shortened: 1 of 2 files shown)


mrubens:
Addressed the review finding: rather than adding `/tmp/.config` to the allowlist, the install/version-check layer now runs under a disposable `HOME=/tmp/opencode-build` that is removed at the end of the same `RUN` — the reviewer's second, more robust option. Any dotdir npm or OpenCode writes (now or in future versions) stays out of the shipped layer, with no allowlist to maintain. Runtime `HOME=/tmp` from `runtime-base` is unaffected since the export is scoped to that one `RUN` shell.

PR #145, opened 2026-07-10
[Feat] Record a death certificate when the OpenCode server exits unexpectedly

daniel-lxs:
PR 2 of the observability pair — #140 (the diagnostic rail) is merged; this is the first producer on it. Replaces #141, which GitHub auto-closed when the stacked base branch was deleted.

## What
The worker keeps a bounded ring (last 50 lines, 300 chars each) of the OpenCode subprocess's output while it runs. If the process exits while the task is **not** being cancelled, it records one durable `opencode_unexpected_exit` diagnostic event: exit code, signal, uptime, output tail, and a `/proc/meminfo` + worker-RSS snapshot taken at the instant of death — plus a Sentry capture so we get alerted. Reconnect-spawned processes are certified individually.

This is the fact that decides the open mystery: exit 137 = killed for memory (the 128 MiB request theory), a panic stack = OpenCode bug, clean exit = something else. The last two investigations couldn't make that distinction because the evidence died with the sandbox.

## Guarantees (per the requirements)
- **Performance:** constant cost per output line (ring append with caps); certification work happens only on unexpected exits.
- **Behavior:** observer-only. The exit hook cannot alter the harness lifecycle — cancellation-driven exits are excluded, callback failures are caught and logged, reconnect behavior is untouched.
- **Privacy:** the output tail flows through the rail's secret scrubbing (keys, bearer tokens, JWTs, credential assignments) and per-line/message caps before persisting; memory snapshot is numbers only.

## Tests
Collector: exit facts + tagged tail, ring/line caps, blank-line and missing-`/proc` safety. Full harnesses + run-task suites: **423 tests pass**; typecheck / eslint / prettier / knip clean.

Not auto-merged — ready for review.

### DIFF
--- a/apps/worker/src/run-task/create-harness.ts
+++ b/apps/worker/src/run-task/create-harness.ts
@@ -14,8 +14,11 @@ import {
 } from '../commands/setup/setup-mcps';
 import type { HarnessLogger } from '../logging';
 
+import { captureWorkerException } from '../monitoring/sentry';
+
 import type { RunTaskCallbacks, RunTaskContext } from './types';
 import { buildHarnessCommandEnv } from './harnesses';
+import { createDiagnosticEventRecorder } from './diagnostic-events';
 import { ReconnectableHarness } from './reconnectable-harness';
 import { subscribeHarnessCallbacks } from './subscribe-harness-callbacks';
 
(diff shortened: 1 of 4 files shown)


daniel-lxs:
Carrying over the known-noise caveat from #141: deliberate harness restarts (e.g. the MCP-reconnect respawn path) terminate opencode without the task-cancel signal, so they will be certified as SIGTERM exits. The killer signatures that matter (exit 137 / SIGKILL / panic stacks) are unambiguous; a follow-up can thread an expected-restart flag to silence the SIGTERM entries.

PR #147, opened 2026-07-10
[Fix] Address review findings on the diagnostic rail and death certificate

daniel-lxs:
Follow-up addressing all outstanding openmote review findings on #140 and #145 (merged before the reviews were read — my miss).

| Finding | Fix |
| --- | --- |
| 🔴 #141/#145-1: deliberate reconnects/restarts certified as unexpected exits + Sentry alert on every routine reconnect | `terminateSubprocess` marks the subprocess (WeakSet) before killing; certification skips marked exits. All deliberate teardown paths funnel through that one method. |
| 🟡 #141-2: death tail silently depended on the URL-wait never detaching its line listeners | Dedicated lifetime readers feed the ring; the `onLine` coupling is removed. |
| 🟡 #140-1: reader returned the *oldest* 500 events, dropping end-of-run diagnostics on busy tasks | Query descending with the limit, reverse to chronological before returning. |
| 🟢 #140-2: caller `details.kind` could override the recorder's classification | `kind` is applied last. |
| 🟢 #140-3: 40-hex git SHAs / digests fully redacted, blunting post-mortem value | Hash-shaped values keep an 8-char identifying prefix (`a1b2c3d4…[redacted]`); real key formats (sk-/ghp_/JWT/assignments) remain fully redacted. |
| 🟢 #145-2: "memory at the instant of death" overstated what's captured | Renamed to `memoryAfterExit` with honest docs: sandbox-wide reading after reclaim, worker RSS is the witness; exit code/signal is the authoritative OOM signal. |

## Tests
Two new (hash-prefix redaction, kind-override guard); full run-task + harnesses suites: **425 pass**; typecheck (worker+web) / eslint / prettier / knip clean.

Waiting for bot review + Dan before merge.

### DIFF
--- a/apps/web/src/trpc/commands/tasks/run-events.ts
+++ b/apps/web/src/trpc/commands/tasks/run-events.ts
@@ -1,4 +1,4 @@
-import { asc, db, eq, taskRunEvents } from '@roomote/db/server';
+import { db, desc, eq, taskRunEvents } from '@roomote/db/server';
 
 const MAX_RUN_EVENTS = 500;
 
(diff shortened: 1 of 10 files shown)


daniel-lxs:
Fixed the 🟡 finding: the expected-exit mark is now gated on the subprocess being alive at kill time (`markExpectedSubprocessExitIfAlive`). A crash that manifests as a disconnect leaves the process already dead when the cleanup path runs, so it is no longer marked and its certificate + Sentry alert fire as intended; deliberate teardowns of live processes stay quiet. Covered by three new unit tests (live → marked; crashed → not marked; already-signalled → not marked).

PR #209, opened 2026-07-11
Add Blaxel sandbox provider

mrubens:
## Summary

- add Blaxel as a hosted sandbox provider across controller dispatch, compute adapters, lifecycle cleanup, usage tracking, OIDC, configuration, and docs
- automatically build the Roomote worker image into a Blaxel-compatible immutable sandbox image during setup, with persisted provisioning progress and retry UI in onboarding and Settings
- handle Blaxel workload readiness with bounded exponential backoff and preserve the worker image mise toolchain when commands execute through the Blaxel sandbox API

## Validation

- live Blaxel task completed successfully through the ngrok deployment; controller reported worker startup in 3.466 seconds
- real dynamic image provisioning produced an immutable sandbox image and a verification sandbox returned Node v22.17.1
- 130 targeted tests passed across compute providers, web provisioning/UI, and setup types
- pnpm lint:fast
- pnpm check-types:fast
- pnpm knip
- git diff --check

## Notes

- Blaxel image provisioning uses the configured registry-qualified Roomote worker image as its source and persists the resulting sandbox/name:version reference as BLAXEL_IMAGE
- explicit process-level BLAXEL_IMAGE values remain supported as manual overrides


mrubens:
Addressed both review findings in 6fb5ae84:

- Blaxel sandbox creation now cleans up after readiness/preview failures and handles the late-resolve abort race.
- Blaxel exec now disables the provider-side timeout for foreground and detached commands so the caller AbortSignal owns the timeout budget.
- Added contract coverage for post-create cleanup, late-abort cleanup, and timeout: 0.

Validated with the focused compute-provider suite, package lint/typecheck, and the full pre-push lint:fast/check-types:fast/knip checks.

PR #231, opened 2026-07-12
Fix manual sleep for Docker tasks

mrubens:
## Summary

- show the Sleep action for all resumable compute providers, including local Docker
- enqueue provider-neutral manual sleep actions through BullMQ
- route Docker and Blaxel tasks to standby while preserving snapshot behavior for hosted snapshot providers
- deduplicate repeated requests and register the queue in Bull Board and graceful shutdown

## Root cause

The menu gated Sleep on snapshot capability and called a snapshot-only mutation. Docker supports resumable sleep through retained-container standby, so it was intentionally excluded despite the runtime already supporting standby resume.

## Validation

- SDK sleep queue tests: 2 passed
- web overflow menu tests: 8 passed
- BullMQ sleep lifecycle tests: 34 passed
- package TypeScript checks: SDK, BullMQ, web
- package ESLint checks: SDK, BullMQ, web
- Prettier and git diff checks

### DIFF
--- a/apps/bullmq/src/index.ts
+++ b/apps/bullmq/src/index.ts
@@ -41,6 +41,7 @@ import { teamsSuggestedTasksOnboardingFollowupJob } from './jobs/teams-suggested
 import { startSnapshotQueue } from './snapshot-queue';
 import { startSlackPrInactivityQueue } from './slack-pr-inactivity-queue';
 import { startPrReviewNotificationQueue } from './pr-review-notification-queue';
+import { startTaskSleepQueue } from './task-sleep-queue';
 
 // Resolve auto-generated auth keypairs before any queue worker starts so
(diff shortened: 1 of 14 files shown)


mrubens:
Addressed both review items in 9cc6c75d:

- Manual sleep now waits for the BullMQ job terminal result. Worker failures and timeouts reject the web mutation, so the existing error toast is shown instead of a false success toast. Duplicate in-flight requests wait on the same job result.
- The web command now validates the raw persisted provider and rejects null before capability checks; it no longer falls back to Docker.

Added queue coverage for terminal waiting, duplicate waiting, and asynchronous worker failure propagation. Focused SDK (3), web (8), and BullMQ (34) tests pass, along with SDK/web type checks and lint.

PR #245, opened 2026-07-12
[Fix] Route Amazon Bedrock API keys through Mantle

mrubens:
## Why this PR exists

Amazon Bedrock keys generated by the current AWS console authenticate against Bedrock Mantle, while Roomote was routing them through OpenCode's legacy Bedrock Runtime provider.

## What changed

- Route Amazon Bedrock models through a custom Anthropic-compatible OpenCode provider at the regional Bedrock Mantle endpoint.
- Preserve the existing `amazon-bedrock` settings identity while migrating legacy Runtime model IDs to `bedrock-mantle/` IDs.
- Apply Anthropic thinking options to Mantle-backed Claude models.
- Label the credential as a Mantle API key and link directly to the AWS key console with matching-region guidance.
- Document the Mantle model IDs and add a patch changeset.

## How it was tested

- Focused Vitest coverage: 144 tests passed across types, worker, web, and database packages.
- `pnpm lint`
- `pnpm check-types`
- Pre-push `pnpm lint:fast`, `pnpm check-types:fast`, and `pnpm knip`
- Local OpenCode probe returned `OPENCODE_MANTLE_OK`.
- Live Roomote task using `bedrock-mantle/anthropic.claude-haiku-4-5` returned `BEDROCK_MANTLE_OK`.
- Verified the updated provider dialog in Chrome.

## Checklist

- [x] The PR title follows the repo convention
- [x] `pnpm lint` and `pnpm check-types` pass locally
- [x] I added tests and included manual validation above
- [x] I removed secrets, tokens, private keys, and customer data from the diff
- [x] I ran `pnpm changeset`

### DIFF
--- a/.changeset/bright-bedrock-mantle.md
+++ b/.changeset/bright-bedrock-mantle.md
@@ -1,0 +1,5 @@
+---
+'@roomote/web': patch
+---
+
+Route Amazon Bedrock API keys through the Mantle endpoint and clarify Mantle key setup in model settings.
--- a/apps/docs/environment-variables.mdx
+++ b/apps/docs/environment-variables.mdx
@@ -179,8 +179,8 @@ as per-task auth tokens or workspace paths.
 | `OPENCODE_API_KEY`                     | Provider key | OpenCode Zen / Go API key.                                                                                                              |
 | `GEMINI_API_KEY`                       | Provider key | Google Gemini API key. Can also be saved from **Settings > Models**.                                                                    |
 | `GOOGLE_GENERATIVE_AI_API_KEY`         | Provider key | Alternate Google/Gemini provider key forwarded when configured or inferred.                                                             |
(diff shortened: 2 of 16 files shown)


mrubens:
Addressed in 1f1cee92.

- Confirmed the models.dev concern and fixed it: `bedrock-mantle/anthropic.*` IDs now resolve through the canonical `anthropic/*` catalog slug, preserving display names, modalities, context windows, and lab pricing. Added resolver and metadata lookup coverage.
- Backward compatibility with persisted Bedrock Runtime IDs is intentionally out of scope for this change. I removed the partial four-model migration map so we do not imply a migration guarantee we do not intend to support. Existing `amazon-bedrock/*` selections must be replaced with the new Mantle models.

Focused tests and the pre-push lint/typecheck/knip suite pass.

PR #247, opened 2026-07-12
[Fix] Materialize Vertex credentials before OpenCode

mrubens:
## Why this PR exists

Pasted Google Vertex service-account JSON reached OpenCode as a literal file path in the primary worker path. Besides breaking Vertex tasks with `ENOENT`, the downstream provider error could echo the credential JSON.

## What changed

- Materialize inline `GOOGLE_APPLICATION_CREDENTIALS` at the shared OpenCode config-generation boundary, before any provider process can observe it.
- Preserve operator-managed credential file paths unchanged.
- Write generated credential files with mode `0600`.
- Fail closed on materialization errors instead of forwarding raw JSON to OpenCode.
- Remove the later, narrower duplicate materializer and add a worker patch changeset.

## How it was tested

- Direct OpenCode Vertex probe with `google-vertex/gemini-3.5-flash` returned the expected response.
- Live Roomote/Modal task reproduced the pre-fix failure; the temporary key was immediately revoked and removed from Roomote after its JSON appeared in the provider error.
- `pnpm exec dotenvx run -f .env.test -- pnpm --filter @roomote/worker exec vitest run src/run-task/agent-home.test.ts src/sandbox-server/lib/harnesses/__tests__/opencode-server-bootstrap.test.ts` (39 tests)
- `pnpm --filter @roomote/worker check-types`
- `pnpm lint:fast`
- `pnpm check-types:fast`
- `pnpm knip`

Post-fix live Roomote verification is pending only the replacement Chrome download being exposed to the local filesystem.

## Checklist

- [x] The PR title follows the repo convention
- [x] Fast lint and type checks pass locally
- [x] I added tests and included manual validation notes
- [x] No secrets, tokens, private keys, or customer data are included in the diff
- [x] Added a changeset


mrubens:
Follow-up diagnosis and fix pushed in a13a76b5.

The first live retest showed that credential materialization succeeded, but `OPENCODE_COMMAND` launches through `bash -lc`; Bash then sourced Roomote’s shared `BASH_ENV` and re-exported the original inline JSON over the generated path. The worker now creates a credential-free OpenCode `BASH_ENV` overlay that sources the shared environment first and then reasserts the materialized credential path.

Validation added/performed:
- Regression test invokes a real `bash -lc` with a shared env file containing inline JSON and confirms the child shell sees only the generated credential path.
- Live Modal task with harmless dummy service-account JSON now reaches OpenSSL PEM parsing (`NO_START_LINE`) instead of treating/echoing JSON as a filesystem path.
- Focused 39-test worker suite, worker typecheck, and pre-push lint/typecheck/knip pass.

Both real keys used during diagnosis were revoked immediately after their JSON appeared in the pre-fix error, removed from Roomote, and deleted locally. Final real-key pass is pending the new Chrome download being moved into the local filesystem.

mrubens:
Addressed the duplication note in ad4bee23. `resolveOpenCodeDataDir` now lives in the shared OpenCode config module and is used by Vertex credential materialization, `auth.json`, and the login-shell `BASH_ENV` overlay. The bootstrap-local copy was removed.

Focused 39-test worker suite, worker typecheck, and the pre-push lint/typecheck/knip gate all pass.

mrubens:
Addressed in 9c1f5670: moved the Google Vertex credential-materializer JSDoc back to immediately precede `materializeInlineGoogleCredentials`, leaving the shared data-directory helper with only its own documentation. Worker lint and the full pre-push lint/typecheck/knip gate pass.

mrubens:
Final live verification passed after the credential file was moved into place:

- Direct Vertex API probe: passed
- Roomote provider save: passed
- Live Modal worker task using Vertex Gemini 3.5 Flash: returned exactly `ROOMOTE_VERTEX_E2E_OK`
- No provider error or credential content appeared in the task output.

PR #309, opened 2026-07-13
Fix compute provisioning tests for the r3 runtime schema bump

daniel-lxs:
## What

develop is red: #301 bumped `WORKER_RUNTIME_SCHEMA_VERSION` 2→3, but its merge CI path-filtered away the `apps/web` test suite, so seven stale test expectations surfaced only when later merges (#291) touched `apps/web` and made CI actually run them.

**No runtime bug** — #301 never touched the provisioning currency logic; verified against its diff. This is test-only:

- "Current" fixtures use `WORKER_RUNTIME_SCHEMA_VERSION` / `WORKER_RUNTIME_SCHEMA_TAG` instead of hardcoded `2` / `-r2`, so future bumps don't break them
- The compute-providers derivation stubs now `importOriginal`, so fixtures and assertions exercise the real templateRef derivations instead of hand-rolled suffix-less strings
- Two tests that kept passing after the bump for the wrong reason (fixtures silently became "stale", flipping them onto untested paths) are re-anchored to the current version with strengthened assertions

## Testing

- Full `apps/web` suite: 281 files / 1969 tests green (was 7 red on develop)
- `check-types:fast`, prettier clean

Unblocks develop and #305.

### DIFF
--- a/apps/web/src/trpc/commands/compute/compute-provisioning.test.ts
+++ b/apps/web/src/trpc/commands/compute/compute-provisioning.test.ts
@@ -1,3 +1,7 @@
+import {
+  WORKER_RUNTIME_SCHEMA_TAG,
+  WORKER_RUNTIME_SCHEMA_VERSION,
+} from '@roomote/types';
 import type {
   SetupNewComputeProvisioningState,
   SetupNewState,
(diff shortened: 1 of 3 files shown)


daniel-lxs:
Superseded: develop absorbed #304 (same test fix via the CI-failure automation) plus the #306/#307 provisioning rework while this was queued, and #310 handles the remaining duplicate-import typecheck error. Closing.

PR #337, opened 2026-07-14
Stream Docker project logs to a well-known file and add them to the Logs panel

mrubens:
## Problem

Docker projects had no standard log. `docker compose up` output went only to the worker's startup logger, container logs lived solely inside dockerd (reachable only via `docker compose logs` from a shell), and the task Logs panel listed just `harness.log` plus detached-command logfiles — nothing for Docker projects.

## Changes

**Worker**
- Each Docker project now gets a well-known log file at `/tmp/roomote-docker-projects/<compose-project>.log`, created up front so the Logs panel's `tail -f` finds it immediately.
- The file receives: a preparing/starting header, redacted Compose startup output, failure diagnostics (`compose ps` + recent logs) when startup fails, and — after a successful `up` — a detached `docker compose logs --follow --no-color --timestamps` stream for the sandbox's lifetime.
- Log writing lives in a separate `docker-project-logs` module with best-effort error handling, so it can never affect project startup (and is cleanly mockable in tests).

**Shared types**
- `toComposeProjectName` moves from the worker into `@roomote/types`, alongside new `getDockerProjectLogFilePath`, so the worker and web derive identical paths.

**Web**
- The Logs sidebar ("Logs in this environment") now lists one entry per configured `docker_project`, tailed over the existing `tailMulti` subscription.

## Testing

- Docker project tests now mock the log module (previously my first cut spawned real processes in tests — caught and fixed) and assert the follower starts with the right compose args.
- Full worker suite (1323 tests), typecheck (types/worker/web), lint, and knip all pass.

### DIFF
--- a/apps/web/src/app/(sandbox)/task/[taskId]/hooks/use-log-files.ts
+++ b/apps/web/src/app/(sandbox)/task/[taskId]/hooks/use-log-files.ts
@@ -1,6 +1,8 @@
 import { useContext, useEffect, useMemo } from 'react';
 import { useStore } from 'zustand';
 
+import { getDockerProjectLogFilePath } from '@roomote/types';
+
 import { useEnvironment } from '@/hooks/environments';
 
 import type { LogfileInfo } from './use-sandbox-store';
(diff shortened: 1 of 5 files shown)


mrubens:
Addressed in the latest commit: failure output appended to the project log is now passed through `redactContainerDiagnostics` with the project's sensitive values — both in `startDockerProject`'s failure path (where `error.message` carries Compose stderr) and in the outer per-project catch (skipped only before resolution succeeds, when no project env has been substituted yet). The rethrown startup error is redacted the same way, so downstream warnings don't leak either. New test drives a compose failure whose message embeds a substituted secret and asserts both the thrown error and every appended log line show `APP_SECRET=[redacted]`.

mrubens:
Addressed in the latest commit: the optional-project failure path no longer hands the caught `Error` to the debug logger (its `cause` chain still carried the raw Compose stderr that `createStartupLogger` would serialize into `harness.log`). Both `userLog.warn` and `debug.error` now log the redacted message string. New test drives an optional project failure with a substituted secret in the Compose error and asserts everything logged or appended shows `APP_SECRET=[redacted]` with no raw value.

PR #403, opened 2026-07-15
feat: keep the sandbox provider step visible when runtime env satisfies it

mrubens:
Stacked on #402 (base branch is `feat/roomote-cloud-compute-provider`; retarget to `develop` after it merges).

## What

Mirrors #388's communication/source-control treatment for the **compute** step, which was the remaining step that silently vanished when env vars preconfigured a provider:

- **`compute-provider` completes only on a genuine choice** (the wizard pick or a persisted Settings default) — runtime satisfaction alone no longer skips it. A deployment that preconfigures a sandbox provider still shows the picker with that provider preselected: one click, nothing to fill in, but the user sees what they're getting and that alternatives exist.
- **`compute-config` skip is scoped to the chosen provider.** Previously the deployment-wide `setupSatisfied` short-circuit meant that on a deployment with a preconfigured provider, choosing an *unconfigured* bring-your-own provider (e.g. E2B) skipped its credential step entirely — the same scoping bug #388 fixed for source control.
- A chosen provider missing from the offered providers list is treated as having nothing to configure (the stale-choice guard already covers exclusions), preserving behavior for under-specified statuses.

No changes to the picker UI or persistence — the existing `saveComputeProviderChoice` flow already persists the choice and short-paths credentialless providers past the config step.

## Tests

Two new `useSetupFlow` cases: the seeded-deployment picker-still-renders scenario, and the chosen-unconfigured-provider-still-configures scenario. Full web client project passes (201 files / 1395 tests); lint and check-types clean.

### DIFF
--- a/apps/web/src/app/(onboarding)/setup/StepComputeConfig.client.test.tsx
+++ b/apps/web/src/app/(onboarding)/setup/StepComputeConfig.client.test.tsx
@@ -236,6 +236,15 @@ describe('StepComputeConfig', () => {
                   setupProvisionable: false,
                 },
                 {
+                  envVarName: 'E2B_TEMPLATE_ID',
+                  label: 'Worker Template',
+                  category: 'infrastructure',
+                  runtimeSatisfied: false,
+                  savedSatisfied: false,
+                  defaultSatisfied: false,
+                  setupProvisionable: true,
+                },
+                {
                   envVarName: 'E2B_DOMAIN',
                   label: 'E2B Domain',
                   required: false,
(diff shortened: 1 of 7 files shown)


mrubens:
Addressed in `fcd8be0c3` — took both halves of the suggestion since they protect different paths:

- **Choice mutation commits when already configured**: `saveComputeProviderChoice` now treats a fully-configured chosen provider like the credentialless path and commits the runtime default in the same transaction — the wizard skips its config step, so the choice is the confirmation.
- **Skip requires the effective default to match**: `compute-config` now skips only when the chosen provider is the effective persisted/runtime default, so a canonical /setup re-entry with pre-fix persisted state (choice recorded, default not committed) renders config once and commits through its confirmation instead of advancing while dispatch targets the old default.

New coverage: command tests for commit-on-configured-choice and no-commit-on-unconfigured-choice; a flow test pinning configured-but-not-default → `compute-config`. The Local Docker fixture now reflects the default the server persists with the choice. Full web client project green (201 files / 1399 tests).

mrubens:
Addressed in `91b47868e`: the hosted worker-image requirement in `StepComputeConfig` is now scoped to when saving would actually derive or provision from it — i.e. a required managed artifact (base image ref / template id / snapshot name / image) is still unsatisfied. The recovery path that re-confirms an already-provisioned provider (saved E2B creds + existing template, locked local-tag `DOCKER_WORKER_IMAGE`) can now complete and commit the dispatch default; fresh hosted setups keep the guard since their pending artifact genuinely needs a registry-qualified image.

Test updates: the fresh-E2B disable test's fixture gains the `E2B_TEMPLATE_ID` managed field it was missing (every real hosted catalog entry carries one — without it the fixture would read as nothing-to-provision), plus a new test pinning the recovery path continuing on a local-tag image. Full web client project green (201 files / 1400 tests).

mrubens:
Addressed in `d14076c83`: `missingHostedWorkerImage` now carries the pending-managed-infrastructure predicate, so the warning, the auto-opened advanced section, and the Continue guard move together — the recovered already-provisioned state neither blocks nor warns, and `hostedRequirementMet` collapses to its negation. The recovery test asserts the warning no longer renders, and the misleading-advanced-toggle fixture now keeps `MODAL_BASE_IMAGE_REF.defaultSatisfied` consistent with its non-hosted-ready worker image (they cannot diverge in production — the default derives from that image).

PR #434, opened 2026-07-16
Clarify Azure DevOps Entra setup

brunobergher:
## Summary
- clarify that an existing Teams Entra app can be reused after adding Azure DevOps access
- show the ADO OAuth redirect URI only for delegated auth setup
- clarify service principal setup copy and the ADO organization slug label

## Tests
- pnpm lint:fast
- pnpm check-types:fast
- pre-push hook: oxlint, residual lint, check-types:fast, knip


brunobergher:
Addressed the outstanding review item by pointing the Entra Enterprise Apps link at the Enterprise applications pane instead of App registrations.

PR #441, opened 2026-07-16
[Fix] Accept Discord button clicks and keep their task threads anchored

daniel-lxs:
Two defects found while dogfooding the routing-confirmation flow against a real Discord bot.

## Every button click was silently discarded
The envelope schema typed `data.id` as a **string**, but Discord sends the pressed component's **numeric layout id** there — only slash commands send a snowflake string, which is exactly why `/link` worked while `All repositories` did nothing at all. Every component interaction failed validation with `invalid_discord_event` (400) and was quarantined to the dead-letter stream. Now both shapes are accepted.

## Confirmed launches detached their task thread
When routing needs a confirmation card, the launch happens on the **button interaction**, which has no message of its own — so the anchor from the original mention was lost and the task thread fell back to detached (visible as "Roomotedev started a thread" instead of a thread on the requester's message). The API now stamps the anchor from the triggering message; it rides along in the pending route and survives the round trip to the confirmed launch.

Interactions genuinely have no anchor of their own, so `/new` and suggestion-card buttons keep their detached threads by design.

## Validation
- api discord suite 61/61, communication discord-event 8/8 (new: numeric `data.id` envelope parses; anchor survives the pending-route round trip)
- check-types 26/26, oxlint, format, knip green
- Verified live: mention → card → click → task launches, thread created, early-title rename, cancel button all working

Found by the same dogfooding session as #432 and #438 — three real Discord-contract bugs the mock harness structurally could not catch, because it encoded our assumptions about Discord rather than Discord's actual behavior.

### DIFF
--- a/apps/api/src/handlers/discord/__tests__/index.test.ts
+++ b/apps/api/src/handlers/discord/__tests__/index.test.ts
@@ -319,6 +319,8 @@ describe('Discord Gateway event handler', () => {
           communicationProvider: 'discord',
           communicationChannelId: 'dm-1',
           communicationMessageId: 'message-1',
+          // A real channel message carries an anchor for its task thread.
+          communicationAnchorMessageId: 'message-1',
         },
       }),
     );
--- a/apps/api/src/handlers/discord/__tests__/routing-confirmation.test.ts
+++ b/apps/api/src/handlers/discord/__tests__/routing-confirmation.test.ts
@@ -158,6 +158,57 @@ describe('Discord routing confirmation', () => {
     );
   });
 
+  it('carries the thread anchor through the pending route to the launch', async () => {
+    // A confirmation card defers the launch to a button interaction, which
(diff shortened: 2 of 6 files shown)


daniel-lxs:
Verified live against real Discord after rebasing onto #438.

Before the fix (task run 51) vs after (task run 52), comparing the launch payload's source message id against the created thread id:

| task_run | source message | thread id | anchored |
| --- | --- | --- | --- |
| 51 (before) | `…690394857704` | `…848897863770` | ❌ |
| 52 (after) | `…925093134387` | `…925093134387` | ✅ |

The thread id now equals the triggering message id, and Discord renders the task thread nested under the requester's message rather than as a detached "Roomotedev started a thread" line. Full path exercised: mention → routing card → button click (previously a silent no-op, now accepted) → task launch → anchored thread → early-title rename.

PR #444, opened 2026-07-16
[Fix] Stop Discord from unfurling Roomote's own task links

daniel-lxs:
## Problem

Every Discord message carrying a task link renders a Roomote marketing embed underneath it — **"Roomote — Cloud management for Roomote."** — because Discord unfurls any link it finds. Roomote posts task links constantly, so the preview is pure noise stapled to a message the reader already understands.

Slack has always sent `unfurl_links: false, unfurl_media: false` on every outbound message (`slack-notifier.ts`, `communication-provider.ts`, `thread-reply-footer-ops.ts`, `router-debug.ts`). Discord unfurling is the divergence, not the other way round.

## Change

Set Discord's `SUPPRESS_EMBEDS` message flag (`1 << 2`) on outbound messages, which is the documented equivalent.

The important subtlety: `SUPPRESS_EMBEDS` hides **every** embed on the message, including ones we send deliberately — Roomote renders images as embeds. So the flag is only set on messages that carry no embeds of their own; an image-bearing chunk is left alone. The chunking loop decides this per batch, since a long message with images splits into a first chunk with embeds and later chunks without.

Applied to the three places that send or edit user-visible text:
- `postMessage` — conditional on the batch carrying no image embeds
- `editMessage` — never carries embeds
- `editInteractionResponse` — never carries embeds

`deferInteraction` is untouched; a deferral has no content, and the follow-up edit is what carries the text.

## Verification

`flags` is accepted on all three endpoints, confirmed against Discord's docs rather than assumed — Create Message and Edit Message document `SUPPRESS_EMBEDS`, and Edit Webhook Message (which backs Edit Original Interaction Response) documents `flags` as "`SUPPRESS_EMBEDS` and `IS_COMPONENTS_V2` only".

166 tests pass in `packages/communication`. New coverage asserts the flag is set on a plain message and on an edit, and — the regression that matters — that a chunk sending image embeds does **not** set it.

### DIFF
--- a/packages/communication/src/__tests__/discord-provider.test.ts
+++ b/packages/communication/src/__tests__/discord-provider.test.ts
@@ -58,6 +58,9 @@ describe('DiscordCommunicationProvider', () => {
       enforce_nonce: true,
       embeds: [{ description: 'Screenshot' }],
     });
+    // SUPPRESS_EMBEDS hides embeds the message carries itself, so a chunk
+    // sending images must not set it.
+    expect(requests[0]?.body).not.toHaveProperty('flags');
     expect(requests[1]?.body).toMatchObject({
       content: 'b'.repeat(25),
       allowed_mentions: { parse: [] },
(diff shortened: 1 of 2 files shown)


daniel-lxs:
Good catch — fixed in d81c2c92, and following it found something worse than the reported case.

`editInteractionResponse` edits a **deferral that may be EPHEMERAL**: `apps/discord-gateway/src/dispatch.ts:155` defers slash commands with `data: { flags: 64 }`, and `apps/api/src/handlers/discord/index.ts:203` answers them with `ephemeral: true`. That set includes `/link`, whose reply carries an account-linking URL. Since editing `flags` rewrites the whole bitfield rather than adding to it, suppressing embeds on that edit risks clearing EPHEMERAL and posting that URL publicly. Discord's docs imply EPHEMERAL is not editable, but that is not a guarantee worth a privacy leak.

So `SUPPRESS_EMBEDS` is now set only while creating a message, and only when the message sends no embeds of its own:

- `postMessage` — kept, conditional per chunk on carrying no image embeds. This is where the unfurls actually appeared.
- `editMessage` — dropped. It has no callers, so the flag there was speculative and only bought the trap you described.
- `editInteractionResponse` — dropped, per the ephemeral risk above. These messages render their links as buttons, so they do not unfurl anyway.

Added a test asserting no PATCH request carries `flags`, so neither can be reintroduced without a failure.

PR #445, opened 2026-07-16
[Fix] Snapshot credentials persist after task sleep

mrubens:
> Opened on behalf of Matt Rubens. [View the task](https://roomote.roomote.ai/task/1vuvfycuv8ur1?utm_source=github-comment&utm_medium=link&utm_campaign=github_pr_review_follow_up) or mention @roomote-roomote for follow-up asks.

## What changed

Worker snapshots now scrub credential material after setup and before sleep or environment snapshots. The sleep-time scrubber resolves OpenCode credential paths from the task runtime home and task runtime environment, including `XDG_DATA_HOME`, so it removes the files the OpenCode bootstrap actually materializes.

## Why this change was made

Snapshot images include the sandbox filesystem. Resolving paths from the worker process could leave task-scoped OpenCode authentication and Google credential files in a snapshot.

## Impact

Snapshot resumes continue to re-materialize credentials from the task payload, while saved snapshot images no longer retain the task-scoped OpenCode credential files.

PR #446, opened 2026-07-16
Request a sandbox-server scrub before queue-triggered snapshots

mrubens:
Follow-up to #445.

#445 scrubs worker-managed credential files before the snapshots the worker itself initiates (the environment-snapshot command and the sleep handoff). The snapshot queue can also snapshot a sandbox whose worker never ran that step — for example the sleep-check recovery paths that snapshot a resumable run after a missed heartbeat.

This adds a `scrubSnapshotSecrets` sandbox-server procedure that runs the same pre-snapshot scrub, and has the BullMQ snapshot job call it best-effort (10s timeout) before asking the compute provider for a filesystem snapshot:

- The scrub is idempotent and everything it removes is re-materialized from the dequeue/resume response at the next run start, so it is requested on every queue-triggered snapshot regardless of trigger path.
- Failures never block the snapshot: no sandbox server URL, an unreachable sandbox server, or an RPC error are recorded as task-run decision events (`pre_snapshot_scrub_skipped` / `pre_snapshot_scrub_unavailable`) and the job proceeds.

Tests cover the new procedure (scrub invoked with the harness logger, errors propagate) and the snapshot job behavior (scrub requested before `createSnapshot`, snapshot continues when the scrub is unavailable, skip path when the run has no sandbox server URL).

### DIFF
--- a/apps/bullmq/src/jobs/snapshot.test.ts
+++ b/apps/bullmq/src/jobs/snapshot.test.ts
@@ -22,6 +22,7 @@ const {
   mockDrainLinearMessagesToResumeRun,
   mockDrainSlackMessagesToResumeRun,
   mockRecordComputeProviderUsage,
+  mockWithSandboxServerRpcClient,
   mockMarkTaskStartParallelCountEndedAt,
   mockSyncTaskStateFromRuns,
   transactionFn,
@@ -57,6 +58,7 @@ const {
     mockDrainLinearMessagesToResumeRun: vi.fn() as AnyMock,
     mockDrainSlackMessagesToResumeRun: vi.fn() as AnyMock,
     mockRecordComputeProviderUsage: vi.fn() as AnyMock,
+    mockWithSandboxServerRpcClient: vi.fn() as AnyMock,
     mockMarkTaskStartParallelCountEndedAt: vi.fn() as AnyMock,
     mockSyncTaskStateFromRuns: vi.fn() as AnyMock,
     transactionFn: vi.fn() as AnyMock,
@@ -145,6 +147,7 @@ vi.mock('@roomote/slack', () => ({
 
 vi.mock('@roomote/sdk/server', () => ({
   recordComputeProviderUsage: mockRecordComputeProviderUsage,
+  withSandboxServerRpcClient: mockWithSandboxServerRpcClient,
 }));
 
 vi.mock('../monitoring/sentry', () => ({
@@ -193,6 +196,7 @@ describe('snapshotJob', () => {
       linearOrganizationId: null,
     });
     mockFindSnapshotBySourceInstance.mockResolvedValue(null);
+    mockWithSandboxServerRpcClient.mockResolvedValue(undefined);
     mockDrainLinearMessagesToResumeRun.mockResolvedValue({ resumed: false });
     mockDrainSlackMessagesToResumeRun.mockResolvedValue({ resumed: false });
   });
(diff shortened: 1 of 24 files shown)


mrubens:
Addressed all three findings in e47908ab9:

1. **Task runtime scrub context** — `createServer` now receives `taskRuntime: { homeDir, runtimeEnv }` from run-task and exposes it on the tRPC context; the `scrubSnapshotSecrets` procedure passes it through to `scrubSandboxSecretsBeforeSnapshot`, so OpenCode credential files under a task-scoped HOME are located the same way the worker's own scrub call locates them.

2. **Partial-failure reporting** — `scrubSandboxSecretsBeforeSnapshot` still attempts every step but now returns the list of failed steps. The mutation throws when any step failed, so the snapshot queue records `pre_snapshot_scrub_unavailable` instead of `pre_snapshot_scrub_completed` for an incomplete scrub.

3. **Quiescing credential writers** — added a credential write barrier (`apps/worker/src/lib/credential-write-barrier.ts`). The scrub engages it and waits for in-flight writes to drain before removing files; once engaged, the source-control token refresh loop skips its refresh writes and `reloadDeploymentEnvVars` rejects instead of rewriting env.sh. The latch is one-way, which matches the sandbox lifecycle: after a snapshot request the sandbox is either torn down or resumed with credentials re-injected at run start.

mrubens:
Addressed the follow-up finding in 68c6df63e: the barrier is no longer a one-way latch.

When the queue's terminal createSnapshot-failure path clears the snapshot request and the sandbox may still be running, it now best-effort calls a new `restoreScrubbedCredentials` sandbox-server procedure, which:

1. releases the credential write barrier,
2. re-materializes the source-control token files via an immediate token refresh (rather than waiting for the loop's next scheduled refresh), and
3. rewrites env.sh / runtime env / harness command env from the deployment's current env vars (shared helper factored out of `reloadDeploymentEnvVars`).

Outcomes are recorded as `post_failure_credential_restore_completed` / `_unavailable` run events and never change the failure handling. The instance-not-running terminal path intentionally skips the restore since there is no surviving sandbox to recover. Harness-managed credential files (e.g. OpenCode auth) are not rewritten: the running harness holds those in memory and any new run start re-materializes them from the dequeue response.

mrubens:
Follow-up in 7b080e2ff: the restore now also rewrites the OpenCode credential files the scrub removed.

The previous restore covered source-control token files and the deployment env but skipped the OpenCode data-dir files on the assumption the running harness held them in memory. That assumption fails for the materialized Google service-account file — Google's auth library re-reads the `GOOGLE_APPLICATION_CREDENTIALS` path when minting tokens — so a surviving Google-backed task would have pointed at a deleted path after a nominally successful restore. `auth.json` has the same shape of problem, since OpenCode persists OAuth token refreshes back to that file.

New `rematerializeOpenCodeCredentialFiles` (in `agent-home.ts`, next to the bootstrap-time materializers) rewrites both files at their bootstrap-time paths from the deployment's current env values, resolving paths with the task runtime HOME/XDG_DATA_HOME while letting fresh deployment values win over the bootstrap-time rewrites. `restoreScrubbedCredentials` calls it after the env reload; any failure surfaces through the existing failed-step error so the queue records `post_failure_credential_restore_unavailable`. Files whose source env value is absent are skipped, and the raw env values never leave the sandbox process (the reload RPC still returns names only).

PR #447, opened 2026-07-16
Add a flag-gated inference gateway so provider keys stay out of task sandboxes

mrubens:
## Summary

Adds an inference gateway at `/api/inference/:provider/*` and, behind the `InferenceGateway` feature flag, routes task-sandbox inference through it so provider API keys stay on the control plane. Every key-authenticated provider in the setup catalog is covered: OpenRouter, Anthropic, OpenAI, Google Gemini, Vercel AI Gateway, Requesty, Baseten, Together AI, Moonshot AI, MiniMax, OpenCode Zen, xAI, and Amazon Bedrock (Mantle, with region-templated upstreams resolved from AWS_REGION per request). Google Vertex (request signing) and ChatGPT subscriptions (client-managed OAuth refresh) still flow directly. Sandboxes authenticate with their run-scoped token (`ROOMOTE_CLOUD_TOKEN`); the gateway resolves the deployment's provider key server-side, injects it upstream, and streams the response back.

This extends the existing broker pattern (integration MCP proxy, git credential proxy, comms brokering, OIDC federation) to LLM traffic. The flag is off by default; nothing changes for deployments that don't enable it.

## The gateway

- **Auth**: registered as `task-token` in the central route-policy manifest; the handler additionally verifies the run token's task run still exists (same pattern as the MCP proxy). The Anthropic and Gemini SDKs send their "API key" through `x-api-key` / `x-goog-api-key` rather than `Authorization`, so the token middleware accepts the run token from those headers on the `/api/inference` surface only.
- **Path allowlists**: each provider exposes exact inference endpoints only (messages/chat/responses/embeddings/models, plus `count_tokens`); Google model routes match nested paths because the model ID and action live below `/models`. Everything else 403s, so a run token can never reach a provider's account, billing, or admin surface (e.g. `/v1/messages/batches` and OpenRouter `/v1/key` are rejected).
- **Key resolution**: server-side with the same precedence the task runtime uses (runtime env first, then persisted encrypted deployment env vars), including the `GOOGLE_GENERATIVE_AI_API_KEY` alias for Gemini. MiniMax and Bedrock Mantle proxy through their Anthropic-compatible endpoints; the aggregators are OpenAI-compatible bearer-auth; Vercel uses the `@ai-sdk/gateway` protocol under `/v1/ai`. Upstream bases follow models.dev, the registry OpenCode itself resolves providers from.
- **Streaming**: requests and responses stream through in both directions using the no-body-timeout dispatcher and logged proxy stream wrapper, so long SSE inference streams work and client disconnects clean up.

## The flag-gated wiring

**Flag**: `InferenceGateway` (deployment metadata key `inference_gateway`), togglable from the admin metadata controls. Evaluation fails closed to direct provider keys if unavailable, so a Redis outage cannot break task inference.

When enabled:
- The dequeue/resume/env-reload path withholds gateway-covered provider keys from the sandbox env (including raw deployment env vars carrying those key names) and emits `R_INFERENCE_GATEWAY_URL` derived from the sandbox-reachable platform API URL. Control-plane inference (routing, titles, summaries) keeps direct keys since it holds no run token.
- Each controller spawn path evaluates the flag and withholds covered operator keys from the worker daemon env.
- The worker's OpenCode config rebases covered providers onto the gateway (`baseURL` = gateway URL + provider + SDK base-path suffix, `apiKey` = `{env:ROOMOTE_CLOUD_TOKEN}`), following the Bedrock Mantle pattern. OpenRouter attribution headers are preserved; the `openai` provider is untouched when a ChatGPT subscription record is present.
- Keys the gateway cannot serve (Vertex service-account JSON) and ChatGPT-subscription OAuth still flow directly; `AWS_REGION` also still flows since it is configuration, not a secret.

Provider definitions (ids, key env vars, upstream URLs, allowlists, SDK base-path suffixes) live in `@roomote/types` so the gateway and worker share one source of truth.

## Testing

- Gateway handler tests: token-type rejection, missing task run, unknown provider, exact-path allowlist (including account-surface paths nested under former prefixes), missing key, header injection/stripping for all three auth styles, query string and SSE passthrough, upstream error handling, method restrictions
- Resolver, OpenCode config, middleware token-extraction, and worker-env tests for the flag-gated behavior
- Full suites pass across `types`, `feature-flags`, `db`, `sdk`, `compute-providers`, `controller`, `api`, and `worker`; typecheck, lint, knip clean
- Verified end-to-end locally (Modal sandbox, OpenRouter): opencode config rebased onto the gateway, no provider key values anywhere in the sandbox filesystem or process envs, task inference streamed through `/api/inference/openrouter/v1` and completed normally

## Rollout notes

- Suggested order: enable the flag on internal deployments (OpenRouter) first.
- Environment-snapshot setup env still carries provider keys when stored as deployment env vars (setup commands may reference them); the pre-snapshot scrub (#445) keeps them out of snapshot images.


mrubens:
@roomote-roomote address the feedback

mrubens:
Addressed the /code-review findings in `fcf06bef9` (also merged `develop`, which fixes the duplicate-import CI failure). Summary:

**Fixed in code**
- **Flag gate on the endpoint** — `/api/inference` now checks the `InferenceGateway` flag and fails closed; it is no longer live on deployments that haven't opted in.
- **Encoded-traversal guard** — `..%2F` / dot-segments are rejected before the nested-path allowlist (google, vercel).
- **Docker/self-host URL** — the sandbox gateway URL is now built worker-side from the container-reachable platform URL, not dequeue's raw `TRPC_URL`.
- **Single served-keys signal (`R_INFERENCE_GATEWAY_KEYS`)** — the resolver withholds/advertises exactly the configured gateway-covered keys; dequeue redaction, the worker harness-env strip, and the provider rebase all key off that one set. This fixes: warm-sandbox daemon-env re-injection, the redaction-stripped-user-keys regression, and the withheld-without-rebase asymmetry for repo-pinned providers.
- **Snapshot env** now goes through the same gateway-aware withholding.
- **Gemini precedence** prefers `GOOGLE_GENERATIVE_AI_API_KEY`; stale Bedrock comments corrected; flag evaluator and Bedrock constants de-duplicated.

**Rollout requirements (documented, not code-enforced)**
- Enable the flag only once all worker sandboxes/snapshots are on a build containing this change — an old worker resuming a snapshot won't understand `R_INFERENCE_GATEWAY_KEYS` and would get neither a key nor a gateway route (same class as the R_* rename rollout).
- Drain warm/standby sandboxes when flipping the flag: a sandbox spawned before the flip keeps provider keys in its daemon `/proc` until recycled (the harness env and `env.sh` are scrubbed, but the daemon's own initial environ can't be rewritten in place).
- Gateway mode requires provider keys reachable by the **API** service (deployment env or DB), not only the controller's process env; the gateway returns a clear 404 otherwise.

mrubens:
@roomote-roomote review this again

PR #509, opened 2026-07-18
[Fix] Tasks keep retrying when provider billing fails

mrubens:
> Opened on behalf of Matt Rubens. [View the task](https://roomote.roomote.ai/task/3cb2ojsp9erj7?utm_source=github-comment&utm_medium=link&utm_campaign=github_pr_review_follow_up) or mention @roomote-roomote for follow-up asks.

## What changed

When a provider reports a billing, suspension, or payment-required failure as a retryable OpenCode session status, the worker now treats it as terminal: it aborts the OpenCode retry loop, surfaces the provider message, clears queued prompts, and ends the task. Message-only retry statuses such as bare `Payment required` (no status code or error code) are classified the same way, matching how `handleSessionStatus` actually invokes the classifier.

Transient provider failures continue to use the existing retry/backoff behavior.

## Why this change was made

Some providers mark account or billing failures as retryable. OpenCode then enters unbounded backoff while the task sits in a retry loop that cannot succeed. Roomote needs to recognize those failures from the status payload it actually receives—including message-only forms—and fail the task cleanly instead.

## Impact

Tasks that hit terminal provider billing errors stop sooner with a clear provider message, instead of spinning on retries. Ordinary transient provider errors still retry as before. No UI changes.

### DIFF
--- a/apps/worker/src/sandbox-server/lib/harnesses/__tests__/opencode-server.test.ts
+++ b/apps/worker/src/sandbox-server/lib/harnesses/__tests__/opencode-server.test.ts
@@ -1989,6 +1989,80 @@ describe('OpenCodeServerHarness', () => {
     }
   });
 
+  it('terminates an OpenCode retry loop for message-only payment required status', async () => {
+    const { client, harness } = createHarness();
+    const taskEvents: TaskEvent[] = [];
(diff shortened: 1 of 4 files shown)


mrubens:
@roomote-roomote address PR comment

PR #566, opened 2026-07-18
feat: support safe inline SVG widgets

daniel-lxs:
## Summary
- allow inline SVG through the widget sanitizer using DOMPurify's SVG profile
- preserve chart primitives such as shapes, paths, text, gradients, and clipping
- strip active or externally loaded SVG markup, including SVG animation elements, filters, embedded HTML, images, and external references
- tell agents that inline SVG is supported while keeping the prompt focused on the sandbox boundary
- cover safe chart markup and malicious SVG payloads with focused tests

## Security boundary
Widgets remain scriptless and network-isolated. The iframe CSP continues to use `default-src 'none'`, while the worker sanitizer removes executable attributes and unsafe SVG elements or references before persistence. Custom CSS behavior is unchanged.

## Validation
- `mise exec -- pnpm --filter @roomote/worker exec vitest run src/mcp/roomote-mcp-server/__tests__/show-widget.test.ts src/mcp/roomote-mcp-server/__tests__/tool-descriptions.test.ts`
- `mise exec -- pnpm --filter @roomote/worker check-types`
- `mise exec -- pnpm --filter @roomote/worker lint`

### DIFF
--- a/apps/worker/src/mcp/roomote-mcp-server/__tests__/show-widget.test.ts
+++ b/apps/worker/src/mcp/roomote-mcp-server/__tests__/show-widget.test.ts
@@ -97,6 +97,56 @@ describe('handleShowWidget', () => {
 });
 
 describe('show-widget helpers', () => {
+  it('keeps static inline SVG chart primitives', async () => {
+    const html = await sanitizeWidgetHtml(`
+      <svg viewBox="0 0 200 100" role="img" aria-label="Bar chart">
+        <defs>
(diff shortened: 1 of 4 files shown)


daniel-lxs:
Addressed the review note in 0e27be69: the tool contract now says only that inline SVG is supported in the existing scriptless, network-blocked sandbox. It no longer claims SVG is static or that CSS animation/filter behavior is stripped; custom CSS behavior remains unchanged.

PR #574, opened 2026-07-19
[Improve] Frame provider comment follow-ups as untrusted content

mrubens:
## Description

Follow-up to #570, which added shared prompt-hardening helpers (`buildMentionRequestBlock`, `buildUntrustedContentPolicy`) and applied them to the GitHub-facing prompts. The non-GitHub provider comment handlers still interpolated the triggering comment into follow-up task messages with plain `> ` quoting and no untrusted-content policy.

The GitLab, Bitbucket, Azure DevOps, and Gitea comment handlers now compose their existing-task follow-up messages the same way as the GitHub issue-mention handler:

- The triggering comment (authored by a commenter who passed the linked-account gate, so it is the request to act on) lands in an escaped `<mention_request>` block instead of `> ` quoting.
- The PR/MR title is escaped with `escapeTaskContextText`, matching the GitHub composition.
- The shared untrusted-content policy is appended to each built follow-up message.
- The now-unused local `formatQuotedText` helpers are removed.

## Test plan

- Handler tests now mock the `@roomote/cloud-agents/server` barrel with distinctive marker fakes and assert the follow-up message routes text through the right builders, mirroring the GitHub handler test update in #570.
- The Bitbucket suite gains its first coverage of the reusable-task follow-up path (steer into an active owner task).
- Ran the four affected `apps/api` handler test directories (21 files, 146 tests), `pnpm lint`, `pnpm check-types`, and `pnpm knip`.

### DIFF
--- a/.changeset/provider-comment-untrusted-framing.md
+++ b/.changeset/provider-comment-untrusted-framing.md
@@ -1,0 +1,5 @@
+---
+"@roomote/web": patch
+---
+
+Apply the untrusted-content prompt framing to GitLab, Bitbucket, Azure DevOps, and Gitea comment follow-up messages, wrapping the triggering comment in a mention-request block and appending the shared injection-resistance policy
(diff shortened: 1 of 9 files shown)


mrubens:
@roomote-roomote resolve conflicts pls

PR #580, opened 2026-07-19
[Feat] Show subscription provider usage on Models settings

mrubens:
## What

Adds a usage line under the ChatGPT, GitHub Copilot, and Kimi for Coding provider rows in **Settings > Models**, showing current plan usage such as remaining Copilot premium requests or percent used of the 5-hour and weekly rate-limit windows, with reset times.

## How

- **`packages/types`**: new normalized `SubscriptionProviderUsage` / `SubscriptionUsageWindow` types plus the provider usage endpoint constants.
- **`packages/db`**: `getSubscriptionProviderUsage()` with one fetcher per provider, reusing credentials the deployment already holds:
  - Copilot: `copilot_internal/user` with the stored OAuth token, reading the `premium_interactions` quota snapshot.
  - ChatGPT: `backend-api/wham/usage` via `getFreshChatGptAccessToken()`, so it rides the existing advisory-locked refresh.
  - Kimi for Coding: the Kimi Code usage endpoint with the configured `KIMI_API_KEY`, with a `/v1/usage` fallback on 404.
- **`apps/web`**: admin-gated `subscriptionUsage.list` tRPC query and the usage line rendering in `InferenceProviderSection` (1 minute stale time). Only normalized numbers reach the client, never tokens.
- **`apps/docs`**: short note on the new usage display in `models.mdx`.

## Notes on the upstream endpoints

None of these usage endpoints are officially documented; each is the endpoint the provider's own CLI polls for its usage display. Live payloads were verified against all three providers and differ in places from what third-party tooling documents (ChatGPT now nests windows under a singular `rate_limit` key; Kimi serializes numbers as strings and uses proto-style enum values). Parsers accept the known field aliases, the live shapes are pinned in tests, and any fetch/parse failure resolves to null so the UI omits the usage line rather than erroring. Window labels are derived from the reported window durations rather than hardcoded, so label text survives upstream window changes.

## Testing

- 12 unit tests covering the parsers (documented and live payload shapes), 404 fallback, not-connected short-circuits, and aggregator failure isolation.
- Verified end-to-end against all three live provider APIs on a connected deployment.
- `pnpm lint`, `pnpm check-types`, `pnpm knip`, and the full `packages/db` lib suite pass.

### DIFF
--- a/apps/docs/models.mdx
+++ b/apps/docs/models.mdx
@@ -137,6 +137,13 @@ list.
 - **Models** controls the provider/model pairs that are available, the default
   model, and specialized model roles.
 
+For subscription-based providers (ChatGPT, GitHub Copilot, and Kimi for
+Coding), the provider row also shows current plan usage when the provider
+reports it, such as remaining Copilot premium requests or the percent used of
+the ChatGPT 5-hour and weekly limits. These numbers come from unofficial
+provider endpoints, so the usage line is hidden whenever the provider does not
+return usable data.
+
 Admins can enable or disable models from the task model list. The default model
 must stay enabled, because Roomote uses it when a task does not request a
 specific model.
--- a/apps/web/src/components/settings/ChatGptConnectDialog.tsx
+++ b/apps/web/src/components/settings/ChatGptConnectDialog.tsx
@@ -87,6 +87,9 @@ export function ChatGptConnectDialog({
           await queryClient.invalidateQueries({
             queryKey: trpc.chatgptSubscription.status.queryKey(),
           });
+          await queryClient.invalidateQueries({
+            queryKey: trpc.subscriptionUsage.list.queryKey(),
+          });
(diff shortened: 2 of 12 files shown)


mrubens:
Addressed the review feedback:

- **CodeQL (URL substring check)**: the aggregator test's mock fetch now matches the hostname exactly instead of using a substring check (a366fd4).
- **Component test mocks**: the `InferenceProviderSection` test's `useTRPC`/`useQuery` mocks now cover the `subscriptionUsage` dependency, plus two new tests for the usage line (rendered under connected rows, hidden on errored rows or missing data) (73a786d).
- **Stale usage after provider changes**: `subscriptionUsage.list` is now invalidated alongside the existing provider/status invalidations: save provider, delete provider, disconnect either subscription, and successful completion of both connect dialogs (73a786d).

PR #627, opened 2026-07-20
[Improve] Validate environment button for Local Docker settings

daniel-lxs:
## What changed

Settings → Sandboxes → Local Docker gains a **Validate environment** button that runs the same checks the controller preflights before a Docker spawn (added in #622):

- Docker daemon reachable
- worker image available locally (with an explicit pull attempt)
- worker release archive present on disk

Results render as a per-check list (pass / fail / skipped with reasons), so operators see everything that needs fixing at once instead of failing fast.

## How it works

The web app has no Docker socket in production Compose, so validation runs as a **BullMQ request/reply job** consumed by the bullmq service (which holds the socket in the `docker-compose.compute-docker.yml` overlay). The tRPC mutation (`compute.validateDockerEnvironment`, admin-gated) enqueues the request and waits on the job result with a 120s bound; if no worker consumes the queue, the UI reports the validation service as unavailable rather than a failed environment.

The check logic lives in `packages/compute-providers` with injectable docker/fs runners. The compose overlay now passes `DOCKER_WORKER_IMAGE` / `DOCKER_WORKER_RELEASE_PATH` to the bullmq service so its checks match what the controller will actually use.

## Why this change was made

#622 made boot failures precise after the fact; this closes the loop by letting self-hosters find the same problems **before** their first task — the moment where opaque failures previously caused churn.

## Validation

- 5 new unit tests for the check logic (163 total in compute-providers pass)
- Live smoke test of the full producer → Redis → worker → Docker round trip against a real daemon: healthy path, unreachable daemon, and unpullable image all classified correctly
- `pnpm lint:fast`, `pnpm check-types:fast`, `pnpm knip` green

### DIFF
--- a/apps/bullmq/src/docker-validation-queue.ts
+++ b/apps/bullmq/src/docker-validation-queue.ts
@@ -1,0 +1,64 @@
+import { Queue, QueueEvents, Worker } from 'bullmq';
+
+import { DOCKER_VALIDATION_QUEUE_NAME } from '@roomote/types';
+import type { DockerEnvironmentValidationResult } from '@roomote/compute-providers';
+
+import { getRedis } from './redis';
+import {
+  type DockerValidationJobData,
+  dockerValidationJob,
+} from './jobs/docker-validation';
+
+export function startDockerValidationQueue() {
+  const connection = getRedis();
+
+  const queue = new Queue<
+    DockerValidationJobData,
+    DockerEnvironmentValidationResult,
+    string
+  >(DOCKER_VALIDATION_QUEUE_NAME, {
+    connection,
+    defaultJobOptions: {
+      // Validation is interactive and single-shot: the settings UI waits on
+      // the result, so a retry would only delay an accurate failure report.
+      attempts: 1,
+      removeOnComplete: { age: 600, count: 50 },
+      removeOnFail: { age: 3600 },
+    },
+  });
+
+  const worker = new Worker<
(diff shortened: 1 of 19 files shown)


daniel-lxs:
On the two review findings:

**Image default in the overlay**: `docker-compose.compute-docker.yml` already sets the identical `${DOCKER_WORKER_IMAGE:-roomote-worker:local}` default on the controller service, and per `deploy/deployment-catalog.json` that overlay is only part of the `self-host-docker-compute` shape — `installer-production` doesn't include it. So within the overlay, bullmq now resolves the image exactly as the controller does; in installer-production both services derive it identically via `resolveEffectiveDockerWorkerImage` (and the web passes its resolved image in the job payload besides). `73fc97ef` additionally mirrors the derivation vars into the bullmq env anchor there so the fallback matches too.

**Socket access**: the raw-socket mount for bullmq in the overlay predates this PR (standby retention uses it), so this change didn't widen that boundary. The finding did surface a real gap though: in `installer-production` bullmq had no Docker access at all, so validation would have falsely reported the daemon unreachable. `73fc97ef` wires bullmq to the existing restricted `docker-proxy` (`DOCKER_HOST=tcp://docker-proxy:2375`, `docker-api` network) — never the raw socket — which the proxy's existing `IMAGES`/`POST` allowances cover for the version/inspect/pull checks.

PR #633, opened 2026-07-20
Notify staging Cloud after develop images

mrubens:
## What changed

- add an independent `notify-staging-cloud` job after successful GHCR publication
- dispatch `roomote-develop-build` to Roomote-Cloud with the exact immutable tag produced by the build
- fail closed unless `ROOMOTE_CLOUD_REPO` is exactly `RooCodeInc/Roomote-Cloud`

## Why

Roomote-Cloud now has a tested staging rollout receiver. This connects successful Roomote `develop` image builds to that receiver while keeping the existing Roomote-Ops notification path unchanged.

## Activation

Before setting `ROOMOTE_CLOUD_REPO`, add a `ROOMOTE_CLOUD_DISPATCH_TOKEN` Actions secret. It should be scoped only to `RooCodeInc/Roomote-Cloud` with repository Contents write permission. Set the repo variable last to activate the job.

## Validation

- YAML parsed successfully
- embedded shell passed `bash -n`
- pre-push oxlint, package lint, fast type checks, and knip passed

### DIFF
--- a/.github/workflows/publish-ghcr.yml
+++ b/.github/workflows/publish-ghcr.yml
@@ -485,3 +485,36 @@ jobs:
             -f "client_payload[image_tag]=${image_tag}" \
             -f "client_payload[sha]=${GITHUB_SHA}" \
             -f "client_payload[ref]=${GITHUB_REF}"
+
+  # Keep staging Cloud delivery independent from the existing Roomote-Ops
+  # production path. This job only follows successful develop image publishes.
+  notify-staging-cloud:
+    name: Notify staging Cloud
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mrubens:
Sequencing guard: blocked on #634. Merge the managed-runtime compatibility first; only then merge this autoship wiring.

mrubens:
Correction to my earlier dependency note: current develop already includes the roomote sandbox/provider, so #634 is obsolete and has been closed. This autoship PR is not blocked on #634, but it should remain draft until we decide whether staging uses the existing direct Modal backend or adds a control-plane lease backend.

PR #637, opened 2026-07-20
[Feat] Sync stored default branches from GitHub repository.edited webhooks

daniel-lxs:
## What changed

- Handle the `repository.edited` GitHub webhook: when the delivery carries a `changes.default_branch` entry, update matching stored repository rows (`sourceControlProvider = github`, matched by `githubRepoId`) to the new default branch. Edits that do not touch the default branch are acknowledged and ignored.
- The App manifest already subscribes to the `repository` event, so existing installations receive these deliveries without any re-configuration.

## Why

Nothing currently refreshes `repositories.default_branch` after the initial installation sync, so a default-branch change on GitHub leaves the row stale until someone manually resyncs the installation. Stale defaults then feed implicit-branch checkouts and branch pickers. Complements #635 (worker-side resilience) and #636 (worker-side self-heal) by fixing the row at the source when webhook delivery is available.

## Validation

- `pnpm exec dotenvx run -f .env.test -- pnpm --filter @roomote/api exec vitest run src/handlers/github/__tests__/handleRepositoryEdited.test.ts` (4 tests: update, non-default-branch edit, missing changes payload, zero-match reporting)
- `pnpm --filter @roomote/api check-types`
- `pnpm lint`

### DIFF
--- a/.changeset/repository-edited-webhook.md
+++ b/.changeset/repository-edited-webhook.md
@@ -1,0 +1,5 @@
+---
+'@roomote/api': patch
+---
+
+Handle the GitHub `repository.edited` webhook so stored repository metadata follows default-branch changes instead of going stale until a manual installation resync.
--- a/apps/api/src/handlers/github/__tests__/handleRepositoryEdited.test.ts
+++ b/apps/api/src/handlers/github/__tests__/handleRepositoryEdited.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+
+const { mockSet, mockWhere, mockReturning } = vi.hoisted(() => ({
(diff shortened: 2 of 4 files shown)


daniel-lxs:
Re the review finding about honoring `isRepoSkipped`: keeping this handler outside the skip guard is deliberate and matches the existing convention in this router. Every `pull_request.*` handler runs its metadata syncs (`syncPrStatus`, `syncPullRequestFact`) before the `isRepoSkipped` guard — the skip policy gates automated processing (reviews, triage, task launches), not stored-metadata accuracy. A default-branch edit updates the same class of deployment metadata: skipping it would leave stale `default_branch` rows for skipped repositories, which then feed branch pickers and implicit checkouts when someone launches a manual task there.

PR #679, opened 2026-07-21
[Improve] Make custom automation report channels optional and thread-anchored

mrubens:
## Summary

Two changes to custom automations, following up on how the built-in channel automations behave:

**1. The report destination channel is now optional.** Previously a channel was required at three layers (UI form, tRPC command, db validation). Now the destination provider select includes a **None** option; without a destination the run happens silently and its results appear only in the task view. The stored `target` uses the existing `{}` column default, typed via a new `OptionalAutomationTarget` + `isConfiguredAutomationTarget` guard. Partially specified targets are still rejected.

**2. When a destination is set, the run is anchored to that channel with real thread semantics.** Custom automation tasks previously launched with only `channels: { slackChannelId }` and a raw user prompt: the agent's `post_to_slack_channel` report was a fire-and-forget top-level message that nothing threaded under, and user replies never routed back to the task. Now the launch mirrors the automation work-item execution-task pattern:

- The payload carries a `customAutomationId` marker plus the Slack `channel`/`slackChannel` fields, so the sandbox gets Slack reply context and registers `send_chat_reply`.
- The Slack `thread_reply` late-bind gate accepts `customAutomationId` alongside `automationWorkItemId`, so the run's first `send_chat_reply` posts a root message in the destination channel and binds it as the task thread via `bindLateSlackThreadToTask`.
- From then on, later task messages continue the thread and user replies in the thread route back into the task (including snapshot resume after completion).
- The prompt appended to the user's automation prompt instructs reporting through `send_chat_reply` with final results, durable blockers, or required input only: no opening acks and no progress updates, matching the execution-task instruction set.

Non-Slack destinations (Teams/Telegram/Discord) already route replies via the `communication*` payload fields, and their reply path degrades to a clean channel post when there is no inbound message to anchor to.

Also updates the failure message for a configured-but-unresolvable destination from "Report destination is not configured." to "Report destination could not be resolved.", and updates the docs page for the optional destination and thread behavior.

## Testing

- `pnpm check-types`, `pnpm lint`, `pnpm knip` all pass
- SDK unit tests cover the anchored prompt (channel tag, `send_chat_reply` instruction, no-progress instruction, payload stamping) and the no-destination launch (raw prompt, no channel bindings, no connectivity check)
- db tests cover creating without a destination and rejecting a partial destination
- Not yet smoke-tested against a live Slack workspace; the Slack late-bind path is verified by code trace against the execution-task flow it reuses

### DIFF
--- a/apps/api/src/handlers/mcp/slack.ts
+++ b/apps/api/src/handlers/mcp/slack.ts
@@ -230,6 +230,15 @@ function getAutomationWorkItemIdFromTaskPayload(
   return typeof value === 'string' && value.trim().length > 0 ? value : null;
 }
 
+function getCustomAutomationIdFromTaskPayload(payload: unknown): string | null {
+  if (!payload || typeof payload !== 'object' || Array.isArray(payload)) {
+    return null;
+  }
+
+  const value = (payload as Record<string, unknown>).customAutomationId;
+  return typeof value === 'string' && value.trim().length > 0 ? value : null;
+}
+
 function normalizeSlackQuoteText(text: string): string {
   return text.replace(/\s+/g, ' ').trim();
 }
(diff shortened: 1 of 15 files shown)


mrubens:
Addressed the review finding in df3ec0b: the worker's turn-satisfaction state now recognizes `customAutomationId` alongside `automationWorkItemId` (shared `isSilentChannelAutomationLaunch` helper), so custom automation runs skip the opening ack and are required to end with a terminal closeout instead. Added worker tests for both Slack and non-Slack custom automation launches.

mrubens:
Fixed in 35ce233: `customAutomationId` is now declared in `sharedTaskPayloadSchema` next to `automationWorkItemId`, so it survives the `taskSpecSchema` parse in dequeue instead of being stripped as an unknown key. Added a schema test asserting the marker and Slack channel context are preserved on StandardTask payloads.

PR #683, opened 2026-07-21
[Fix] Use the custom automation's name in the report footer and drop automation framing from reports

mrubens:
## Summary

Two polish items from real-workspace testing of custom automation report threads (follow-up to #679/#682):

**1. Footer said "Created via the custom automation automation".** The root message was labeled correctly at creation (#682), but `refreshTrackedAutomationThreadRootFooter` rewrites the footer after replies and resolved its label from the task's initiator key through the built-in automation registry. `custom_automation` has no registry descriptor, so it fell back to the raw key with underscores replaced. The refresh path now resolves the `customAutomationId` payload marker to the automation row and uses the automation's own name as the label, matching the create path.

**2. Reports narrated the automation instead of the result.** Runs were posting messages like "Automation requested X. Outcome: delivered Y." The channel-anchored prompt now instructs the agent to write the report as the result itself, like a teammate sharing what they found or did, without mentioning the automation, the schedule, or that anything requested the work. The footer already carries the attribution.

## Testing

- SDK test asserts the anchored prompt carries the no-automation-framing instruction.
- `pnpm check-types`, `pnpm lint`, `pnpm knip` pass.
- The footer-refresh label path is exercised in a real workspace by replying in a custom automation report thread and checking the root footer text.


mrubens:
Addressed in ab0bf20: the footer-label lookup now scans the task's runs oldest first and takes `customAutomationId` from whichever payload carries it, instead of reading one arbitrary run that could be a resume row without the marker.

PR #689, opened 2026-07-22
[Feat] Add the broker backend so managed deployments hold no Modal credentials

mrubens:
## Problem

The managed `roomote` provider authenticates to Modal with a shared workspace token pair injected by Roomote Cloud into every tenant deployment. One compromised deployment exposes the entire Modal workspace.

## What this adds

A new `broker` value for `ROOMOTE_CLOUD_BACKENDS`: sandbox operations route through the hosting operator's compute broker over HMAC-signed HTTP (timestamp + nonce + method + path + body-hash, per request). In broker mode:

- `ROOMOTE_CLOUD_TOKEN_ID/SECRET` carry the deployment id + a derived per-tenant broker credential — **no Modal workspace credentials exist in the deployment**.
- Registry/ECR pull secrets move entirely broker-side; the factory's broker branch never reads them.
- `ROOMOTE_CLOUD_BROKER_URL` (new env var, deployment-managed and reserved like the rest of the `ROOMOTE_CLOUD_*` set) selects the backend alongside `ROOMOTE_CLOUD_BACKEND=broker`.

`RoomoteBrokerClient` (`packages/compute-providers/src/adapters/roomote-broker.ts`) implements the existing `ComputeProviderClient` interface with Modal-parity semantics: aggregated exec output delivery, the 1s detached-exit grace period, background `onExit` delivery with exec-status polling recovery if the stream drops, snapshot via 202+poll (snapshot still terminates the source sandbox), abort → `AbortError`, and `commandId: undefined`. Non-spawn consumers (bullmq snapshot/sleep-check, sandbox OIDC priming, cleanup) get the broker client for free through the factory.

**Backward compatibility**: the default backend stays `modal`; existing deployments are untouched until Cloud flips their env. `modal` ↔ `broker` is id-compatible — the broker fronts the same Modal workspace with the same `roomote-<slug>` app naming, so machine ids and snapshots stay valid across the flip (doc note added in `roomote-cloud.ts`).

Companion control-plane PR (broker service + provisioning contract): RooCodeInc/Roomote-Cloud#32. This PR must be in the pinned app image before any tenant is flipped to broker mode.

## Verification

- 15 new adapter unit tests (signing incl. base-path prefixes, streaming aggregation, detached grace/onExit, snapshot polling, error mapping, abort) + broker factory tests; full compute-providers (179), types (643), and controller (151) suites pass.
- Cross-repo e2e over real HTTP against the real broker app: 31/31 checks (tenant isolation, revocation, forged credentials, image allowlist).
- Live-Modal e2e with the real workspace token through the broker: create/exec/writeFiles/snapshot→resume/destroy on real sandboxes, including filesystem persistence across snapshot-resume and cross-tenant isolation.

### DIFF
--- a/apps/controller/src/RoomoteController.ts
+++ b/apps/controller/src/RoomoteController.ts
@@ -90,10 +90,19 @@ export class RoomoteController extends BaseController {
       // own spawn function here after the backend resolution.
       case 'modal':
       case 'roomote': {
-        if (provider === 'roomote') {
-          // Throws on an unsupported backend; the resolved value is 'modal'
-          // until more ROOMOTE_CLOUD_BACKENDS exist.
-          resolveRoomoteCloudBackend(resolvedEnv);
+        // Throws on an unsupported backend. `broker` routes all sandbox
+        // operations through the hosting operator's compute broker; the
+        // token pair then carries the tenant id + derived broker credential.
+        const backend =
+          provider === 'roomote'
+            ? resolveRoomoteCloudBackend(resolvedEnv)
+            : ('modal' as const);
+        const brokerUrl = resolvedEnv.ROOMOTE_CLOUD_BROKER_URL;
+
+        if (backend === 'broker' && !brokerUrl) {
+          throw new Error(
+            'ROOMOTE_CLOUD_BROKER_URL is required to spawn broker-backed Roomote Cloud workers',
+          );
         }
 
         const modalTokenId =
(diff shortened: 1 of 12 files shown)


mrubens:
Fixed in 7170a14. Two changes: `consumeExecEvents` now preserves its partial outcome across stream drops (`tolerateStreamErrors` for the background watcher) instead of discarding it, and the poll fallback uses the exec id from whichever consumer observed the `started` event — `streamed.execId ?? grace.execId` — so a late `started` followed by a drop still recovers the exit code. Added a fake-timer regression test for exactly this sequence (no events in the grace window → `started` lands late → stream errors → exit code recovered via status polling and `onExit` fires).

PR #721, opened 2026-07-23
Revert #719: re-land Discord/Telegram buttons and auto-handle as separate PRs

daniel-lxs:
Reverts the behavior of #719 (f96ccef6). The two features it bundled — Discord/Telegram notification buttons and the auto-handle mode — will re-land as separate PRs so auto-handle can be tested in isolation before it ships.

The `task_pull_requests.auto_handle_feedback_by_user_id` column and migration 0020 are intentionally kept: they are additive and dormant, and removing an already-applied migration would desync the drizzle journal on environments that ran it.


daniel-lxs:
Not reverting — testing auto-handle containment on the merged code instead.

PR #722, opened 2026-07-23
[Feat] Resolve bare issue references (#234, ENG-512) against configured repositories

daniel-lxs:
**Draft until reviewed — do not merge without Dan's call.** Rebased onto `develop` after #718 and #728 landed; now two commits: bare-reference resolution + a deadline-dispatch fix its new test caught.

## What changed

Closes the last gap in the lookup gate: the routing precheck already extracts references like `#234`, `acme/web#42`, or `ENG-512` into its `externalReference` field, but nothing acted on it when the message had no pasted URL. Now, when the precheck returns `needsExternalLookup=true` and URL parsing finds nothing, `gatherExternalIssueContext` resolves the reference deterministically:

- `ENG-512`-style identifiers → Linear `get_issue` directly.
- `owner/repo#N` → fetched only if that repo is in the deployment's configured `repositoryNames` (case-insensitive match, configured casing wins).
- Bare `#N` / `N` → fans out across the **first 5** configured repositories (environment order; issue numbers that don't exist in a repo drop out at fetch time); each result is labeled per repo (`[acme/web#234]`) so the informed re-route call disambiguates. Smarter candidate selection (routing memory) is a planned follow-up.

Bare references normalize to canonical URLs and run through #718's registry, so fetch-attempt construction (incl. GitHub's `issue_read` → `get_issue` fallback) stays single-source. Pasted URLs always take precedence over the precheck reference.

## Also in this PR (found by its own test)

A fake-timer test for the shared 8s deadline exposed a pre-existing bug from #693: `withExternalLookupTimeout` received an already-invoked promise, so an exhausted deadline still **dispatched** the next MCP call and merely never awaited it. The helper now takes a thunk and checks the remaining budget before invoking.

## Security posture unchanged

The model proposes reference *text* only; fetch targets come from a closed candidate set (the deployment's own configured repos / the connected Linear workspace), execution stays in server code under the shared 8s fail-open deadline. No model changes — same routing model as develop.

## Tests

105 router tests green: five bare-reference cases (bare Linear id, multi-repo fan-out with per-repo labels, repo-qualified validation incl. unconfigured-repo rejection, seven-repo fan-out capped at five, URL precedence), the deadline-sharing fake-timer test, and an end-to-end `Check issue #234` → fetch → `phase: 'mcp'` case. `pnpm knip`, `tsc --noEmit`, `format:check` clean.

### DIFF
--- a/packages/cloud-agents/src/server/router/__tests__/external-issue-context.test.ts
+++ b/packages/cloud-agents/src/server/router/__tests__/external-issue-context.test.ts
@@ -100,6 +100,25 @@ describe('gatherExternalIssueContext', () => {
     expect(result).toEqual({ contextMessages: [], toolsUsed: [] });
   });
 
+  it('shares one lookup deadline across fetch attempts instead of extending it', async () => {
+    vi.useFakeTimers();
+
+    try {
+      vi.mocked(callRouterMcpTool).mockReturnValue(new Promise(() => {}));
+
+      const pending = gatherExternalIssueContext(
+        createContext('Investigate https://github.com/acme/web/issues/42'),
+      );
+
+      await vi.advanceTimersByTimeAsync(8_000);
+
+      expect(await pending).toEqual({ contextMessages: [], toolsUsed: [] });
+      expect(callRouterMcpTool).toHaveBeenCalledTimes(1);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
   it('resolves a bare Linear identifier from the precheck', async () => {
     vi.mocked(callRouterMcpTool).mockResolvedValue({
       title: 'Fix OAuth callback',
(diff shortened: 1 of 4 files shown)


daniel-lxs:
Added the fake-timer test for the shared 8s deadline — and it caught a real (pre-existing, since #693) bug: `withExternalLookupTimeout` received an already-invoked promise, so once the deadline was exhausted the next MCP fetch attempt was still **dispatched** and merely never awaited. The deadline gated waiting, not dispatching. Fixed in e89da82e by making the helper take a thunk and check the remaining budget before invoking. Net behavior change: a hung first fetch no longer fires a second fire-and-forget MCP call after the budget expires. 105 router tests green.

PR #760, opened 2026-07-24
[Fix] Setup sync repeats after source-control callback

mrubens:
## What changed

Setup URL updates now share one pending query state, so step navigation and provider metadata keep each other's query parameters. Source-control OAuth callbacks also remove their one-time `sync=1` marker through that same path before advancing the wizard.

## Why this change was made

The router applies URL changes asynchronously. A direct callback URL update could leave the shared pending query stale, allowing the next navigation to restore `sync=1` and trigger another repository sync.

## Impact

The setup wizard and its server-rendered documentation remain on the same step after source-control callbacks. Returning to the source-control connect step no longer reprocesses an already-consumed OAuth sync marker.


mrubens:
@roomote-roomote address the feedback

PR #793, opened 2026-07-26
[Fix] Use enumerated tool denials so structured output keeps working

mrubens:
## Summary
- #790 restricted non-task OpenCode sessions with a blanket permission denial plus a wildcard session ruleset. Both also strip the internal mechanism OpenCode uses to fulfil `format: json_schema`, so every structured non-task call (all router surfaces) failed with "Model did not produce structured output" while plain-text calls (titles, summaries) kept working — which is why routing broke on nightly.
- Replace both layers with enumerated per-tool denials via a shared `NON_TASK_TOOL_PERMISSION_DENIALS` constant: object-form `permission` config on server spawn and a matching per-session ruleset. Every real tool stays denied; the structured-output mechanism stays available.
- Operator-supplied per-tool permission entries outside the enumerated set are preserved; our denials win on overlap. A non-object operator `permission` (e.g. blanket `"allow"`) is still discarded.
- Regression tests pin the enumerated shape at both layers so a future "simplification" back to a wildcard fails CI.

Bisected against a live OpenCode 1.17 server: blanket config denial and wildcard session rule each reproduce the failure independently; the enumerated form passes structured and text calls end-to-end through `runNonTaskSdkPrompt` while tools remain denied.

## Test plan
- [x] End-to-end repro against a live OpenCode server: structured + text calls succeed with this change; both wildcard forms fail
- [x] Updated unit tests for config merge (including preserving unenumerated operator entries) and session ruleset shape
- [x] Full `@roomote/cloud-agents` suite (577 tests)
- [x] `pnpm lint`, `pnpm check-types`, `pnpm knip`

### DIFF
--- a/packages/cloud-agents/src/server/__tests__/non-task-provider-usage.test.ts
+++ b/packages/cloud-agents/src/server/__tests__/non-task-provider-usage.test.ts
@@ -368,6 +368,18 @@ describe('resolveOpenCodeSmallModel', () => {
     expect(sessionPermissions.every((rule) => rule.permission !== '*')).toBe(
       true,
     );
+    // The per-prompt tool filter is the fail-closed layer for tools the
+    // enumerated denials cannot name (MCP/plugin tools on external servers):
+    // everything off, with OpenCode's internal StructuredOutput tool as the
+    // only exception so `format: json_schema` keeps working.
+    const promptTools = sessionPromptMock.mock.calls[0]?.[0]?.tools as Record<
+      string,
+      boolean
+    >;
+    expect(promptTools).toEqual({ '*': false, StructuredOutput: true });
+    expect(
+      Object.entries(promptTools).filter(([, enabled]) => enabled),
+    ).toEqual([['StructuredOutput', true]]);
     expect(sessionPromptMock).toHaveBeenCalledWith(
       expect.objectContaining({
         sessionID: 'session-1',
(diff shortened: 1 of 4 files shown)


mrubens:
Addressed in d9b7411a: config content forwarded to the managed helper servers is now allowlisted to model/provider selection keys (`model`, `small_model`, `provider`, `enabled_providers`, `disabled_providers`), with `permission` set to exactly the enumerated denials. Tool-introducing keys (`mcp`, `plugin`, `agent`/`mode`, `tools`) and all operator permission entries are dropped, restoring fail-closed behavior for unknown tools on the managed-server path. Externally configured servers (`OPENCODE_SDK_SERVER_URL`) own their config; the session-ruleset comment now documents that boundary. Verified structured + text calls end-to-end against a live server with the allowlisted config.

mrubens:
Addressed in fcbe1644: every non-task prompt now passes a per-prompt `tools` filter — `{ "*": false, "StructuredOutput": true }` — which disables all registered tools, including MCP/plugin tools on externally configured servers the enumerated permission denials cannot name. The single exception is OpenCode's internal `StructuredOutput` tool, which `format: json_schema` requires; a rename of that tool in a future OpenCode release makes structured calls fail loudly rather than making any tool executable, so the filter fails closed. Verified live: structured and text calls succeed with the filter active, and a prompt explicitly requesting file writes executes zero tools even with no permission rules configured.

PR #808, opened 2026-07-27
fix: stop skipping GHCR publishes for docs-only pushes

mrubens:
## Problem

The Publish GHCR Images workflow skipped all build jobs for docs-only develop/main pushes. That skip is wrong in two ways:

1. **It ships stale content.** Since #754 the app image copies `apps/docs` into the image — the web server reads setup-flow docs from `../docs` at request time — so a "docs-only" change produces a genuinely different app image. The skip predates that change. The classifier also treated every `*.md`/`*.mdx` in the repo as docs, even though some markdown (for example workflow `SKILL.md` sources) is runtime content.

2. **It breaks immutable-tag consumers.** A skipped push leaves the commit with no `<channel>-<sha8>` image tag. `pnpm dev` derives `ghcr.io/roocodeinc/roomote-worker:develop-<first 8 chars of origin/develop>` for hosted compute, so after a changelog-only push lands on develop (for example `acab6738`, where every build/publish job was skipped), Modal sandbox creation fails with an empty "Image build for im-... failed" error and every local task run fails to spawn until the next code push. Verified against the registry: `develop-acab6738` returns 404 while `develop-4a08f555` and the `develop` alias return 200.

## Fix

Remove the docs-only skip entirely: delete the `Classify changed paths` step and the `docs_only` gates on `deployment-acceptance`, `build`, and `create-github-release`. Every develop/main push now builds and publishes, restoring the invariant that every branch commit has a published immutable tag and keeping shipped docs current.

Alternatives considered and dropped:

- **Client-side registry check in `pnpm dev`** (probe GHCR, walk back to the newest published tag): works, but adds machinery to every consumer and leaves the tag gap in place.
- **Retagging the previous image under the new sha for docs-only pushes**: fixes the tag gap cheaply, but publishes a stale app image under the new commit's tag, since docs-only changes do affect the app image (caught by review on an earlier revision of this PR).

Tradeoff: changelog-only and docs-only pushes now pay for a full image build. Those pushes are infrequent, and the previous behavior was silently shipping stale docs, so the skip was not a safe optimization to keep.

## Validation

- Workflow YAML parses cleanly; the diff is a pure removal (1 insertion, 54 deletions).
- No `docs_only` references remain in the workflow.

### DIFF
--- a/.github/workflows/publish-ghcr.yml
+++ b/.github/workflows/publish-ghcr.yml
@@ -31,7 +31,6 @@ jobs:
     permissions:
       contents: read
     outputs:
-      docs_only: ${{ steps.changed-paths.outputs.docs_only }}
       owner: ${{ steps.vars.outputs.owner }}
       version: ${{ steps.vars.outputs.version }}
       product_version: ${{ steps.vars.outputs.product_version }}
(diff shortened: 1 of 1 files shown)


mrubens:
Good catch — confirmed: since #754 the app Dockerfile copies apps/docs into the image (the web server reads setup-flow docs from ../docs at request time), so retagging the previous app image under a docs-only commit would ship stale docs. Rather than restricting the retag, the PR now removes the docs-only skip entirely: the skip predates #754 and was already shipping stale docs by never publishing docs-only changes at all. Every develop/main push now builds and publishes.

PR #811, opened 2026-07-27
fix: allow operator env vars in MCP config substitution; warn on refused references

mrubens:
## Problem

Environment-config MCP servers can reference deployment env vars with `${VAR}` syntax, but the substitution guard filtered by **name shape**: any variable ending in `_SECRET` or `_PRIVATE_KEY` was silently refused, and the literal `${VAR}` text was passed to the MCP server with no warning at any layer. So this natural config just doesn't authenticate, with nothing to go on:

```yaml
mcpServers:
  reddit:
    command: npx
    args: [-y, reddit-mcp-buddy]
    env:
      REDDIT_CLIENT_SECRET: ${REDDIT_CLIENT_SECRET}   # silently left literal
```

The name-shape filter also protected nothing for these vars: operator-defined deployment env vars are already injected into the sandbox runtime env, readable by the agent and every process in it. Refusing to substitute them into MCP config was friction without security.

## Fix

Key the guard off **source** instead of name shape:

- **Operator-defined deployment env vars always substitute**, whatever they're named. On a name collision with a reserved runtime name (e.g. an operator's own `DATABASE_URL`), the operator's value wins — internal values are never reachable.
- **Reserved Roomote runtime names** (`ROOMOTE_*`, `AUTH_TOKEN`, `JOB_AUTH_*`, `PREVIEW_AUTH_*`, `DATABASE_URL`, `REDIS_URL`, `BASH_ENV`) remain non-injectable. This is the boundary that matters: MCP config substitution runs worker-side and could otherwise re-expose values the sandbox-secrets scrubbing deliberately keeps away from tasks.
- The `_SECRET`/`_PRIVATE_KEY` suffix heuristics are gone.

And make any refusal loud instead of silent: a reference to a reserved name warns with guidance, and a reference to an undefined name warns as a likely typo. Unresolved references are still passed through literally (streamable-http headers rely on this for the OpenCode-level `{env:...}` conversion).

Also extracts the reference-scanning regex into `collectEnvVarReferences` next to `substituteEnvVars` so the two can't drift.

## Tests

Covers: operator var with `_SECRET` name substitutes; non-reserved secret-like task env var substitutes; reserved name refused + warned; operator value wins reserved-name collisions; undefined reference warns; fully-resolved config does not warn. Full worker run-task + commands suites pass; lint/typecheck/knip pass.

### DIFF
--- a/apps/worker/src/commands/resume.ts
+++ b/apps/worker/src/commands/resume.ts
@@ -41,6 +41,7 @@ export async function resume(runId: number): Promise<boolean> {
     }) => buildWorkspaceConfig({ environmentId, repo, selectedRepositories }),
     runFn: async ({
       jobContext,
+      userEnvVars,
       workspace,
       workspacePath,
       usesSharedWorkspaceRoot,
@@ -88,6 +89,7 @@ export async function resume(runId: number): Promise<boolean> {
       return runTask({
         ...jobContext,
         envVars: jobContext.envVars,
+        userEnvVars,
         workspacePath,
         prompt: '',
         harnessInstructions: jobContext.harnessInstructions,
--- a/apps/worker/src/commands/run.ts
+++ b/apps/worker/src/commands/run.ts
@@ -49,6 +49,7 @@ export async function run({
       }),
     runFn: async ({
       jobContext,
+      userEnvVars,
       workspace,
       workspacePath,
       usesSharedWorkspaceRoot,
@@ -65,6 +66,7 @@ export async function run({
       return runTask({
         ...jobContext,
         envVars: jobContext.envVars,
+        userEnvVars,
         workspacePath,
         prompt: jobContext.prompt,
         harnessInstructions: jobContext.harnessInstructions,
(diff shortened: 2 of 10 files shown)


mrubens:
Addressed in c4773d53: the operator overlay is now built from the pre-injection `userEnvVars` snapshot (taken in `executeTaskRun` before `injectEnvVars` mutates the env map) and threaded through `runFn` → `runTask` → `createHarness`. As defense in depth, Roomote-namespaced names (`ROOMOTE_*`, `AUTH_TOKEN`, `BASH_ENV`, `JOB_AUTH_*`, `PREVIEW_AUTH_*`) are excluded from the overlay even if they appear in the operator map, so a runtime-injected entry can never be reclassified as operator-provided — while generic reserved names (`DATABASE_URL`, `REDIS_URL`) can still be shadowed by an operator's own values. Added a regression test for the `${ROOMOTE_AUTH_BYPASS_VALUE}` case.

PR #849, opened 2026-07-27
[Fix] Stop Slack progress replies from ending the agent turn

daniel-lxs:
## Problem

In `buildSlackMessageInstructions`, the `slack_response_delivery` block placed an unscoped rule directly under the one that assigns `send_chat_reply` to acks, progress updates, and closeouts:

> Do not run more tools first. The only non-reply exception is `tool_search` when the needed Slack reply/post tool is not visible.

The intent is ordering — post the lifecycle reply *before* doing more work. With nothing scoping it, the rule reads as a general instruction to stop calling tools once the reply is sent.

That combines badly with the `slack_turn_lifecycle` cadence rule, which requires a progress reply rather than letting Slack go silent for more than ten minutes during active work. The resulting loop on Slack-surface tasks:

1. Agent works for a short burst.
2. Decides a `progress` reply is due.
3. Posts a sentence describing what it is about to do next.
4. Ends the turn without doing it.
5. Run goes idle until the user prompts again, then repeats.

The visible symptom is a long-running task that keeps announcing intent and stalling, with each burst of real work shrinking, and the user having to prompt repeatedly to get any forward motion.

`slack_visibility_contract` already says `ack`, `progress`, and `clarification` replies "keep the Slack turn open," so the wrapper contradicted itself. The runtime Slack silence hook in `slack-silence-hook-script.ts` carries the same instruction but ends with "After sending the Slack update, continue the work you were doing" — the prompt-side copy was missing that continuation clause.

## Change

Prompt text only, scoped to the Slack wrapper:

- Rewrite the ordering rule to say what it means — send the due lifecycle reply before running further tools — and note that it orders the reply ahead of more work rather than capping the work that follows.
- Add a rule stating that `ack`, `progress`, and `clarification` replies do not end the turn, and that work continues in the same turn afterward.
- Add a rule stating that only a `closeout` reply or an explicit user pause ends the turn, and that a reply describing what happens next is a `progress` reply that must be followed by doing it.

No behavioral code paths change, and no other surface wrapper carries this rule.

## Validation

- `slackAppMention.test.ts` — 25 passed, updated to assert the new rules and that the old unscoped sentence is gone
- Full `src/server/workflows` suite — 42 files, 246 tests passed
- `pnpm lint` and `pnpm check-types` clean

Prompt-wording changes like this can only be confirmed end to end by watching a real Slack-surface task run past a progress reply; that has not been observed yet and is the thing to check after this lands.


daniel-lxs:
Good catch — fixed in 0db5acef.

You're right that the runtime contract is narrower than what I wrote. The `send_chat_reply` tool description says ack and progress keep the turn open, and separately that "a clarification also ends the turn when the next step depends on the user's answer." My rules folded `clarification` into the same bucket as ack and progress and then added a must-follow-through obligation on top, which would have told the agent to keep working after asking a question it was genuinely blocked on.

Split into three rules now:

- `ack` and `progress` never end the turn. A `clarification` behaves the same way *only while real progress is still possible without the answer*.
- The turn ends on a `closeout`, on a `clarification` whose answer the next step genuinely depends on, or on an explicit user pause. Blocking clarifications are named as a real stopping point — wait for the answer rather than guessing, and don't append a separate "waiting on your answer" message.
- The follow-through obligation is scoped to "when nothing is blocking the next step," so it can no longer override a blocked clarification.

That keeps the original fix intact for the failure this PR targets — a `progress` reply that describes the next step and then ends the turn without taking it — while leaving the blocked-clarification exit exactly as the tool contract defines it.

Workflow suite still green: 42 files, 246 tests. Lint and check-types clean.

daniel-lxs:
Fixed in 170e8f5e — and you were right that this one belongs in this PR, not a follow-up.

I'd noticed the `slack_visibility_contract` looseness earlier and written it off as pre-existing. That was the wrong call: before this PR, line 68 was merely imprecise and nothing contradicted it. Adding a precise blocking-clarification exit in `slack_response_delivery` is what turned it into two rules in the same prompt disagreeing about exactly the case where the agent most needs to stop and wait. My change created the conflict, so it's mine to resolve.

Both visibility contracts now read the same way as the response-delivery rules: `ack` and `progress` always keep the turn open, a `clarification` does so only while real progress is still possible without the answer, and a blocking `clarification` ends the turn.

One thing beyond the letter of the finding, flagged rather than slipped in: `buildChatProviderMessageInstructions` (the generic Teams/Telegram/Discord builder, line 208) carried the identical sentence. That surface does **not** have the turn-ending defect this PR fixes — it has no "do not run more tools first" rule — but it does carry the same disagreement with the `send_chat_reply` tool description. Correcting one visibility contract and not the other would have relocated the contradiction rather than removed it, so both got the same one-line scoping. Happy to split that out if you'd rather keep this PR strictly Slack-only.

Workflow suite green: 42 files, 246 tests. Lint and check-types clean. Full CI passed on the prior commit and is re-running now.

daniel-lxs (commented):
Looks good—no blocking findings. The lifecycle wording now makes the ordering requirement explicit without turning progress replies into stopping points, while preserving blocking clarifications as valid turn exits. The prompt contract is consistent with the runtime hooks, and the focused coverage passes. The remaining live Slack verification noted in the PR is the right follow-up.

PR #893, opened 2026-07-30
[Fix] Destroy compute sandboxes when task runs are canceled

mrubens:
## Problem

Canceling a task run never destroyed its compute-provider sandbox. Canceled runs leave the sleep-check candidate set (it only sweeps active statuses), and none of the cancel writers called `destroyInstance`, so the machine kept running until the provider TTL reaped it, counting against the deployment's concurrent-sandbox capacity the whole time. Enough canceled runs in a TTL window and new tasks fail to start with "The tenant sandbox limit was reached."

## Fix

New shared helper `destroyCanceledTaskRunSandbox` (packages/sdk): re-reads the run and destroys its machine only when the run is actually canceled, has a `machineId`, was not preserved via snapshot, and has no final `compute_provider_usage` record yet. That last guard makes it idempotent against sleep-check, which destroys the instance itself before finalizing. It records the usual `destroy_instance` mutation events and the `compute_provider_usage` `destroy` row, builds the vendor client the same way sleep-check does, and never throws (cancel finalization must not fail because teardown did).

Wired into every cancel path:

- `finishRun` with status `Canceled` (including stop-requested failures normalized to canceled). Runs last so a worker-driven terminate, which awaits this RPC from inside the sandbox being destroyed, gets all other side effects first.
- The direct-cancel writers that never reach `finishRun`: the API cancel handler and the web cancel command.
- The pre-sandbox direct cancel in the stop-task flow (usually a no-op, covers a machine stamped without a reachable sandbox server).
- A post-spawn controller check covering the race where a cancel lands mid-provision, after the run row is terminal but before/while the machine is stamped.

Soft stops are untouched: they never set `Canceled`, so resumable sessions keep their sandbox.

Also maps the compute broker's `sandbox_limit_reached` / "The tenant sandbox limit was reached." error to actionable capacity copy in the web app instead of the raw broker message.

## Tests

- New unit tests for the helper (destroy, skip, idempotency, and failure paths).
- `finishRun` tests: destroys on `Canceled` and on stop-normalized failures; never on `Completed`/`Failed`/`Idle`.
- Stop-task test asserting the direct-cancel path invokes the teardown helper.
- Error-display tests for the sandbox-limit mapping.
- Full sdk task-runs suite, sleep-check, and BaseController suites pass; `check-types` clean for sdk, api, controller, and web.

### DIFF
--- a/apps/api/src/handlers/tasks/__tests__/task-stop.test.ts
+++ b/apps/api/src/handlers/tasks/__tests__/task-stop.test.ts
@@ -14,6 +14,7 @@ const {
   mockTxReturning,
   mockMarkTaskStartParallelCountEndedAt,
   mockCancelTaskRunDirect,
+  mockDestroyCanceledTaskRunSandbox,
 } = vi.hoisted(() => {
   const mockTxReturning = vi.fn();
   const mockTxUpdateWhere = vi.fn(() => ({ returning: mockTxReturning }));
@@ -36,11 +37,13 @@ const {
     mockTxReturning,
     mockMarkTaskStartParallelCountEndedAt: vi.fn(),
     mockCancelTaskRunDirect: vi.fn(),
+    mockDestroyCanceledTaskRunSandbox: vi.fn(() => Promise.resolve('skipped')),
   };
 });
 
 vi.mock('@roomote/sdk/server', () => ({
   withSandboxServerRpcClient: mockWithSandboxServerRpcClient,
+  destroyCanceledTaskRunSandbox: mockDestroyCanceledTaskRunSandbox,
 }));
 
 vi.mock('@roomote/db/server', () => ({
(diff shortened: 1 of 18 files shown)


mrubens:
Addressed the teardown race in 7be2c76d: both destroyers (cancel finalization and sleep-check's `destroyInstanceWithAudit`) now take an atomic redis `SET NX` claim keyed by provider + machine id before the provider delete. The loser sees the claim held and skips without recording a failed mutation; the claim is released when the delete fails so a later attempt can retry, and a redis outage falls back to unserialized teardown rather than leaking the machine. Added a concurrent-caller test (exactly one provider delete, no false failed mutation) plus held-claim and claim-release tests on both the helper and the sleep-check path. Same commit also fixes the CI test failure: the web task-runs command test now mocks `@roomote/sdk/server` so the suite stops pulling schema tables into its hand-rolled db mock.

mrubens:
Addressed the lease-ownership issue in 4a47ac0d. The claim now stores a unique per-caller token; release and renewal run as Lua scripts conditional on that token, so a caller whose lease lapsed can never delete a successor's claim. The lease renews every 5 minutes in the background while destroyInstance is in flight and is settled explicitly: finish() on success (stops renewal, lets the key expire so the residual TTL keeps guarding against a lagging duplicate) and release() on failure (token-conditional delete so retries can re-claim). The renewal timer is unref'd, and the compute client is constructed before claiming so an early failure can't leave an armed lease. Tests cover token-conditional release, renewal during a slow provider delete, and renewal stopping once the claim settles.

PR #935, opened 2026-07-31
[Fix] Sandbox pushes fail when a repo hook needs yarn

pridemusvaire:
## The bug

A repository whose git hooks shell out to `yarn` cannot push from a Roomote sandbox. The worker image installs node, npm, and pnpm through mise but never yarn, so the hook exits with `yarn: not found` — and agents report that to the user as a missing Git credential.

Confirmed on the current worker image:

```
$ docker run --rm ghcr.io/roocodeinc/roomote-worker:develop yarn --version
yarn: command not found
```

## The fix

Worker setup probes for a working yarn and, when it is missing, enables it through Corepack before continuing:

```
corepack enable yarn && corepack prepare yarn@stable --activate && mise reshim
```

Three details that matter:

- **`enable` is scoped to yarn.** Unscoped `corepack enable` also writes `pnpm`/`pnpx` shims next to the resolved corepack binary, which on this image is inside the mise shim directory, so the following reshim publishes them ahead of the mise-managed pnpm. Measured below.
- **The download is timeout-bounded.** `corepack prepare` reaches the registry and this runs inside the serial setup block that gates task start, so a throttled or unreachable registry degrades to a warning instead of hanging the boot.
- **Success is verified, not assumed.** A shim on PATH is not the same as a working binary, so the step re-runs `yarn --version` afterwards and warns if yarn is still not executable. Failure is soft throughout: yarn is a convenience for repositories that need it, not a boot requirement.

## Scoping to yarn is load-bearing

Measured in disposable containers from `ghcr.io/roocodeinc/roomote-worker:develop`:

| | `pnpm --version` | `mise which pnpm` | `yarn --version` | extra shims |
|---|---|---|---|---|
| baseline | 10.29.3 | `installs/pnpm/10.29.3/pnpm` | not found | — |
| `corepack enable` (unscoped) | **11.18.0** | **`installs/node/22.17.1/bin/pnpm`** | 4.18.0 | `pnpx` |
| `corepack enable yarn` | 10.29.3 | `installs/pnpm/10.29.3/pnpm` | 4.18.0 | — |

The unscoped form does not merely add a competing shim, it wins: every sandbox silently moves from pnpm 10.29.3 to 11.18.0, and `pnpm --version` then prints `! Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-11.18.0.tgz` before answering. That would add a registry round trip to every `pnpm install` and fail outright under restricted egress. Scoping to yarn leaves pnpm and npm exactly as the image baked them.

## Test plan

- [x] `apps/worker` setup tests: 14 for this file (7 before), 216 across `src/commands`
- [x] Covers the real failure shapes: `execa` reports a missing binary as `exitCode: undefined` (ENOENT), not `1`, so both that and the throwing-probe path are exercised alongside the soft-failure branches
- [x] Mutation-checked: unscoped `corepack enable`, `corepack enable yarn pnpm`, a removed timeout, `?? 1` → `?? 0`, `mise reshim <tool>`, `&&` → `;`, a swallowed warning, and a `catch` returning true all turn the suite red
- [x] Live image: literal script exits 0, is idempotent across repeated runs, completes in ~1s against the 120s bound, and leaves `npm` at 10.9.2
- [x] `pnpm knip`, `oxlint`, `oxfmt --check`, `tsc --noEmit` clean


pridemusvaire:
Narrowed this PR to the single bug it is named for, and recording what review turned up.

**Scope change.** The delivery-skill change that was bundled here (always `git push --no-verify`) has been removed. It disables a gate on every agent push across every repository Roomote touches, which is a policy decision rather than part of fixing a missing binary, and it should not ride inside a bug fix. It will be proposed separately. This PR is now `apps/worker` only.

**The scoping bug, measured.** The original unscoped `corepack enable` also plants `pnpm`/`pnpx` shims. Measured in disposable containers from the current worker image, that is worse than a competing shim, because it wins:

| | `pnpm --version` | `mise which pnpm` | extra shims |
|---|---|---|---|
| baseline | 10.29.3 | `installs/pnpm/10.29.3/pnpm` | none |
| `corepack enable` | **11.18.0** | **`installs/node/22.17.1/bin/pnpm`** | `pnpx` |
| `corepack enable yarn` | 10.29.3 | `installs/pnpm/10.29.3/pnpm` | none |

Every sandbox would have silently moved to pnpm 11, with `pnpm --version` emitting `! Corepack is about to download .../pnpm-11.18.0.tgz` before answering. Since the corepack step runs on every cold boot (the baked image has no yarn), fixing yarn would have broken the primary package manager everywhere.

**Other findings folded in:**

- `mise reshim` takes no positional argument (`Usage: mise reshim [OPTIONS]`), and a tool name is silently ignored rather than rejected: `mise reshim nodejs` exits 0 and reshims everything. So the original `(mise reshim nodejs || mise reshim node || mise reshim)` chain was not broken, it was redundant, with the second and third arms unreachable. Collapsed to the bare form.
- `corepack prepare` downloads from the registry inside the serial block that gates task start, and had no timeout. Now bounded at 120s, degrading to the existing soft warning. The `yarn --version` probe is bounded too, since it becomes a Corepack entrypoint once the shims exist.
- `execa` reports a missing binary as `exitCode: undefined` (ENOENT), not `1`. Every "yarn missing" test modelled it as `1`, so the branch that actually fires in production was never exercised and could be inverted without turning the suite red. Both that and the throwing-probe path are covered now, and the check reads `result.exitCode === 0` rather than relying on a defaulted value.
- `vi.clearAllMocks()` does not drain queued `mockResolvedValueOnce` values, so a stray value passes in the test that queued it and fails unrelated ones instead. Switched to `resetAllMocks()`.
- The `knip.ts` allowlist entries added earlier were unnecessary. Knip never detected those binaries and was emitting `Remove from ignoreBinaries` hints for both, so they are reverted and that file no longer appears in this diff.

mrubens (approved):
Thanks!

PR #958, opened 2026-07-31
Rename yarn changeset to match PR #935 final scope

pridemusvaire:
## Summary

Follow-up to #935. That PR was narrowed to yarn-on-PATH only (delivery-skill `--no-verify` work was dropped), but the changeset filename still said `push-hooks`. Rename to `sandbox-yarn-on-path.md`. Content is unchanged.

## Why draft

This is pure archaeology: filename only, no runtime or release behavior change. Holding as draft so it does not clutter the ready-to-review queue; convert to ready when a quiet window opens or when someone cares about the stale name for release notes archaeology.

## Test plan

- [x] Rename only; no runtime code change
- [x] Changeset body still accurate for the worker patch


pridemusvaire:
Not needed. Filename-only archaeology with no runtime or release impact; changeset body already matches the yarn-on-PATH fix from #935. Closing.

PR #1018, opened 2026-08-03
[Fix] ACA Sandboxes wake-loop fix and improved suspend state management

tebieshi:
## Related issue

N/A

## Why this PR exists

- [x] A maintainer explicitly invited this PR ~~in the linked issue or discussion~~
- [ ] I am a maintainer / this is internal Roomote work

## What changed

Fixes an Azure Container Apps Sandboxes sleep/wake death loop: after a task was suspended (standby), every wake attempt died on "waking up" and the sandbox never stayed suspended.

Four interlocking defects, verified live against sandbox + DB event timelines:

1. **Zombie workers killed every wake.** A suspended sandbox resumes with its worker process frozen mid-run, but the run is finalized (Completed) at suspend time, so the worker's run token is dead. The zombie kept running, held sandbox-server port 4200, and the next wake's worker died ~1 minute into boot — then the 2-minute stale-worker sweep suspended the sandbox mid-boot, and the cycle repeated on every wake.
   → Workers now terminate themselves when the server reports their run as finalized (heartbeat rejects with the terminal run-guard message), and resume launches reap leftover workers (`pkill`) before booting the new one.
2. **Retained-but-Running sandboxes were invisible.** Sleep-check candidacy requires `snapshotId IS NULL`, so a retained standby sandbox left Running (failed wake, portal stop/start, legacy port wake) was never re-suspended.
   → Standby-retention (5-min cadence) now re-suspends azure retained handles whose instance reports `running`, guarded so it never touches a sandbox an active run owns.
3. **Ports woke sandboxes out-of-band.** `OnDemand` activation resumes a suspended sandbox on any inbound traffic, but standby runs are always Completed at suspend time, so traffic-driven wakes could only orphan.
   → All Roomote-added sandbox ports now use `Manual` activation; wake is deliberate-only (wake prompt / message send).
4. **Deleted sandboxes retried forever.** A 404 from `getInstanceStatus` (sandbox deleted out-of-band) went down the generic retry path and re-failed every minute indefinitely.
   → Definitive not-found now finalizes the run (idle → complete; otherwise fail/cancel).

## How it was tested

- New unit tests: sleep-check not-found finalization (404 → finalize, idle-complete, transient-error retry preserved), standby-retention orphan re-suspend (5 tests incl. active-session and genuine-snapshot guards), worker heartbeat terminal-run exit; azure contract test updated to `Manual` port activation.
- Suites: compute-providers 210/210, bullmq 146/146, worker heartbeat 7/7; `pnpm lint` and `pnpm check-types` pass on all four touched packages.
- Live validation on a production deployment after deploying this branch: wake button works (clean boot on the previously looping sandbox), task suspended to `Stopped` at the keepalive deadline, and the standby handle was recorded for the next wake.

## Checklist

- [x] The PR title follows the repo convention: `[Fix]`, `[Feat]`, `[Improve]`, `[Refactor]`, `[Docs]`, or `[Chore]` followed by a user-facing description
- [x] `pnpm lint` and `pnpm check-types` pass locally
- [x] I added tests or included a clear manual validation note above
- [x] I removed secrets, tokens, private keys, and customer data from code, logs, and screenshots
- [x] If this change should appear in the changelog, I ran `pnpm changeset`

### DIFF
--- a/.changeset/aca-sandboxes-sleep-wake-remediation.md
+++ b/.changeset/aca-sandboxes-sleep-wake-remediation.md
@@ -1,0 +1,5 @@
+---
+'@roomote/web': patch
+---
+
+Fix an Azure Container Apps Sandboxes sleep/wake loop where waking a suspended task repeatedly died on "waking up" and sandboxes never stayed suspended. Workers revived from a suspended sandbox now terminate themselves once the server reports their run as finalized (instead of idling as zombies holding the sandbox-server port), resume launches reap leftover workers before booting the new one, retained standby sandboxes found Running are re-suspended automatically, runs whose sandboxes were deleted are finalized instead of retried forever, and sandbox ports use Manual activation so inbound traffic no longer wakes sandboxes out-of-band.
--- a/apps/bullmq/src/jobs/snapshot.test.ts
+++ b/apps/bullmq/src/jobs/snapshot.test.ts
@@ -33,6 +33,7 @@ const {
   setFn,
   updateFn,
 } = vi.hoisted(() => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- suppressed for oxlint; ESLint's own rule is offloaded and reports this directive as unused, which is a false positive
   type AnyMock = Mock<(...args: any[]) => any>;
 
   const andFn: AnyMock = vi.fn(() => 'and-condition');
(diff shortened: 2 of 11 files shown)


mrubens (approved):
Thanks!

PR #1139, opened 2026-08-06
[Fix] Cap Node heap by default in the shared app image

mrubens:
## Problem

The shared app image launches web, api, controller, and bullmq with no `--max-old-space-size`. V8 sizes its default heap ceiling from the memory it can see, which in a container is the host's total rather than the container's limit. With that much perceived headroom it defers full garbage collection for hours, which causes two problems:

1. **Memory drift on usage-billed hosts.** Measured over 72 hours on two representative deployments, every Node service has a post-GC working set of roughly 300-400 MB but drifts far above it between full GCs (api up to 4.3 GB, web 2.8 GB, bullmq 1.9 GB, controller 1.7 GB). On platforms that bill memory by usage, that drift is billed directly.
2. **Crash loops under container memory caps.** A container memory limit is enforced by the runtime, but nothing tells V8, so the process is OOM-killed before V8 ever feels heap pressure. Reported by a self-hoster running bullmq under a 1 GB cap: sawtooth memory growth and periodic OOM restarts even with no tasks running.

## Change

`.docker/app/entrypoint.sh` now applies an explicit `--max-old-space-size` to each long-running Node service:

| Service | Default heap cap |
|---|---|
| web, api | 768 MB |
| controller, bullmq | 512 MB |

- When a cgroup v2 memory limit is readable (`/sys/fs/cgroup/memory.max`), the default is lowered (never raised) to ~75% of the container's memory, leaving headroom for native memory outside the V8 heap.
- An operator-provided `--max-old-space-size` in `NODE_OPTIONS` always wins, and unrelated `NODE_OPTIONS` flags are preserved (the cap is appended, never clobbered).
- `preview-proxy` (non-Node) and `db-migrate` (short-lived) are unchanged.

Documented in `deploy/railway/README.md` and the self-hosting docs, including the manual `NODE_OPTIONS=--max-old-space-size=<MB>` override, which is also the workaround for images published before this change.

## Testing

- `sh -n` passes on the modified entrypoint.
- The helper was exercised standalone against: no cgroup file, `max` (unlimited), empty file, 1 GB / 512 MB / 128 MB / 8 GB limits, the near-int64 no-limit value (no arithmetic overflow), an operator-set `--max-old-space-size` (wins untouched), and pre-existing unrelated `NODE_OPTIONS` flags (preserved with the cap appended). A 512 MB limit derives a 384 MB heap; an 8 GB limit keeps the service default.

## Notes for review

- The cgroup-derived value only ever lowers the service default. An operator who wants a larger heap on a big container sets `NODE_OPTIONS` explicitly.
- 75% of the container limit was chosen to leave room for native allocations (buffers, code, connections) outside the V8 heap.

### DIFF
--- a/.docker/app/entrypoint.sh
+++ b/.docker/app/entrypoint.sh
@@ -61,17 +61,29 @@ apply_node_heap_cap() {
   # cgroup v1 reports "unlimited" as a page-rounded near-int64 number; the
   # derived_mb < cap_mb comparison below already treats it as no limit.
   self_v2="$(sed -n 's/^0:://p' /proc/self/cgroup 2>/dev/null | head -n 1)"
-  self_v1="$(grep -E '^[0-9]+:([^:]*,)?memory(,[^:]*)?:' /proc/self/cgroup 2>/dev/null | head -n 1 | cut -d: -f3)"
 
   limit_bytes=""
   if [ -n "$self_v2" ] && [ -r "/sys/fs/cgroup${self_v2}/memory.max" ]; then
     limit_bytes="$(cat "/sys/fs/cgroup${self_v2}/memory.max")"
   elif [ -r /sys/fs/cgroup/memory.max ]; then
     limit_bytes="$(cat /sys/fs/cgroup/memory.max)"
-  elif [ -n "$self_v1" ] && [ -r "/sys/fs/cgroup/memory${self_v1}/memory.limit_in_bytes" ]; then
-    limit_bytes="$(cat "/sys/fs/cgroup/memory${self_v1}/memory.limit_in_bytes")"
-  elif [ -r /sys/fs/cgroup/memory/memory.limit_in_bytes ]; then
-    limit_bytes="$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)"
+  else
+    self_v1="$(grep -E '^[0-9]+:([^:]*,)?memory(,[^:]*)?:' /proc/self/cgroup 2>/dev/null | head -n 1 | cut -d: -f3)"
+    # The v1 memory hierarchy can be co-mounted with other controllers at a
+    # combined path, so resolve its mount point from /proc/self/mountinfo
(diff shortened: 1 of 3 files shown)


mrubens:
Addressed in 16500e7: the helper now falls back to cgroup v1's `memory/memory.limit_in_bytes` when v2's `memory.max` is absent, so a v1-constrained container gets the derived ~75% cap. The v1 unlimited sentinel needs no special case — the derived value only ever lowers the cap when it is smaller than the service default, so the near-int64 sentinel keeps the default. Verified standalone: v1 512 MB limit derives 384 MB, v1 unlimited keeps the default, and v2 takes precedence when both files exist.

mrubens:
Addressed in 3947975: the helper now resolves the process's own cgroup path from `/proc/self/cgroup` and reads the limit there first, for both v2 (`0::<path>`) and v1 (the memory controller entry, including co-mounted controller lists like `cpuacct,memory`), falling back to the mount-root files for namespaced mounts. Verified against eight fake layouts, including the flagged one: v1 hierarchy with an unlimited root and a 512 MB limit under the process's path now derives a 384 MB cap. Namespaced v1/v2, self-unlimited, missing `/proc/self/cgroup`, and no-cgroup layouts all keep the intended behavior.

mrubens:
Addressed in 6ae58ec: the v1 memory mount point is now resolved from `/proc/self/mountinfo` (a `cgroup` filesystem whose super options include `memory`), so co-mounted hierarchies like `cpuacct,memory` resolve to their combined path. The conventional `/sys/fs/cgroup/memory` remains only as a fallback when mountinfo is unreadable, and the awk substitution is guarded so a missing mountinfo cannot abort the entrypoint under `set -e`. Verified with the image's own mawk 1.3.4: the exact program picks the co-mounted memory mount out of a multi-mount table, and the full helper passes eight layout scenarios (co-mounted path, standard hierarchy, no mountinfo, namespaced mount, multi-mount table, v2 precedence, no cgroup, operator override).

PR #1147, opened 2026-08-07
[Feat] Deployment-scoped custom MCP servers (remote, OAuth, and stdio)

### LINKED ISSUE #1142: [Feature]: Globally configurable MCP for custom integrations
### Preflight

- [x] I searched existing issues and docs before opening this.
- [x] This proposal is grounded in a real Roomote use case, not a broad generated feature list.

### Problem

I want my tasks on my self-hosted instance to be able to use custom MCP Servers to interface with Services not available as a native Integration

### Current workaround

None

### Proposed behavior

Under Settings -> Integration create a “Custom” integration, which allows to configure custrom MCP servers, which become available in all tasks.

### Affected surfaces

MCP

### Self-hosting implications

Unknown

### Alternatives considered

_No response_

### Documentation impact

_No response_

mrubens:
# [Feat] Deployment-scoped custom MCP servers (remote + OAuth + stdio)

Closes #1142.

Adds a **Custom MCP Servers** section to Settings → Integrations where deployment admins can connect MCP servers that are not in the built-in catalog. Custom servers become available to agents in every task.

## What's supported

- **Remote servers (streamable HTTP)** with no auth, static header auth, or OAuth. Remote servers are delivered to sandboxes as authenticated Roomote proxy URLs (`/api/mcp/custom/<id>`); the proxy injects credentials per request, so header values and OAuth tokens never enter sandbox-readable config or the browser.
- **OAuth** per the MCP authorization spec: RFC 8414/9728 discovery, RFC 7591 Dynamic Client Registration with PKCE, manual client id/secret fallback for servers without DCR, RFC 8707 resource indicators (with a per-server opt-out), server-side token refresh, and an explicit reconnect state when a refresh is definitively rejected.
- **Local (stdio) servers**, launched inside the task sandbox via the same pipeline as environment-scoped `mcpServers` (inheriting the reserved-env-reference defenses and the mise shim fix). Env values are encrypted at rest and write-only in the UI.
- **Per-tool deny lists**, enforced at the proxy (tools/call blocked, tools/list filtered), managed from a tool dialog that lists the server's live tools.

## Security

- New `safeFetch` egress guard for all control-plane traffic to operator-supplied URLs (the proxy data path and every OAuth discovery hop, including URLs the remote server itself supplies): private/special-use ranges blocked, DNS answers vetted and pinned (anti-rebinding), redirects refused. Self-hosters can allow specific internal ranges with `R_CUSTOM_MCP_ALLOWED_PRIVATE_CIDRS` (a CIDR list, deliberately not a boolean).
- Reserved names (`roomote`, catalog ids, `github`, `slack`) and reserved runtime env references are rejected at save time.
- Header names the proxy owns (`authorization`, `host`, `mcp-session-id`, ...) cannot be overridden; header values are validated against CR/LF injection.
- Editing a server's URL or auth mode clears stored OAuth tokens, so credentials can never be replayed against a different endpoint.
- Custom-server secrets are per-value encrypted at rest; list endpoints return key names only; the decrypted stdio config endpoint is restricted to run tokens.
- `R_CUSTOM_MCP_DISABLED` kill switch, independent of `R_CURATED_INTEGRATIONS_DISABLED` (operators who disable the catalog are the primary custom-server audience).

## Notes

- Deployment servers merge after environment-scoped `mcpServers`; the more specific scope wins name collisions.
- Custom server names feed the self-setup catalog so the setup agent stops recommending services that are already wired.
- Docs: new `integrations/custom-mcp-servers` page plus updates to the integrations index and the add-integration skill.
- Rollout note: workers baked into old sandbox snapshots don't fetch custom servers; the feature appears on fresh-snapshot tasks.
- Migration is additive (N-1 safe). OAuth connections reuse `mcpConnections` under a `custom:<serverId>` id namespace, which previous-release code ignores.

## Testing

- Schema validation, CRUD (encryption round-trip, blank-keeps-existing, token clearing), SSRF guard units (blocked ranges, IPv4-mapped bypasses, DNS pinning against a live local server, redirect refusal), proxy integration tests against a mock MCP server (header injection, OAuth, deny lists, redirect refusal, egress guard, body cap), worker merge/redaction tests, SDK delivery and run-token gating tests, and UI client tests.
- Verified end to end on a dev deployment: server CRUD in Settings, live tool listing through the guard, per-tool disables, and an agent task calling the tools through the proxy.

## Screenshots

Custom servers render as ordinary integration cards in the same Connected/Configured grids as the built-in catalog, marked with a small **Custom** badge. A prompt bar at the top is the entry point for adding one.

Custom MCP servers as integration cards

Adding a server starts from one dialog that chooses between a remote (HTTP) and a local (stdio) server:

Add custom MCP server dialog

| Header auth (values write-only) | OAuth (DCR + manual client fallback) |
| --- | --- |
| Edit dialog with header auth | Edit dialog with OAuth |

| Local (stdio) config | Per-tool deny list (proxy-enforced) |
| --- | --- |
| Local stdio edit dialog | Manage tools deny list |

### DIFF
--- a/.agents/skills/add-integration/SKILL.md
+++ b/.agents/skills/add-integration/SKILL.md
@@ -86,7 +86,7 @@ Stop and ask the user to choose one of these paths:
 
 - Build a native MCP server inside Roomote with `serverMode: 'native'` and `connectionMode: 'admin_configured'`
 - Wait for the provider to ship an official remote MCP server
-- Recommend that the user configure a custom MCP server in their environment instead
+- Recommend that the user configure a deployment-wide custom MCP server (Settings → Integrations → Custom MCP Servers; supports header auth and OAuth via the proxy) or an environment-scoped `mcpServers` entry instead
 
 Do not quietly convert a missing remote MCP into a built-in integration without that product decision.
 
--- a/apps/api/src/handlers/mcp/__tests__/custom-mcp.test.ts
+++ b/apps/api/src/handlers/mcp/__tests__/custom-mcp.test.ts
@@ -208,6 +208,20 @@ describe('createCustomMcpProxy', () => {
     expect(lastUpstreamHeaders?.authorization).toBeUndefined();
   });
 
+  it('sends an operator-configured static Authorization header upstream', async () => {
+    mockFindCustomServer.mockResolvedValue(
+      buildServerRow({
+        url: upstreamUrl(),
+        headers: { Authorization: 'enc(Bearer operator-api-key)' },
+      }),
+    );
+
+    const response = await postMcp(createApp(), initializeRequest);
+
+    expect(response.status).toBe(200);
+    expect(lastUpstreamHeaders?.authorization).toBe('Bearer operator-api-key');
+  });
+
   it('injects the OAuth access token for oauth servers', async () => {
     mockFindCustomServer.mockResolvedValue(
       buildServerRow({ url: upstreamUrl(), authType: 'oauth', headers: null }),
--- a/apps/api/src/handlers/mcp/__tests__/custom-mcp.test.ts
+++ b/apps/api/src/handlers/mcp/__tests__/custom-mcp.test.ts
@@ -1,0 +1,323 @@
+import { createServer, type IncomingHttpHeaders, type Server } from 'node:http';
+import type { AddressInfo } from 'node:net';
+
+import { Hono } from 'hono';
+import type { RunTokenContext } from '@roomote/types';
+
+import type { Variables } from '../../../types';
+
+const {
+  mockEnv,
+  mockFindTaskRun,
+  mockFindCustomServer,
+  mockFindConnection,
+  mockGetValidAccessToken,
+} = vi.hoisted(() => ({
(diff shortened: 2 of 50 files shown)


mrubens:
Fixed in 97537bb. The cap was enforced after `c.req.json()` had already buffered and parsed the body (and `JSON.stringify` allocated it a second time), so the limit protected nothing.

Capped upstreams now read through `readBoundedRequestText`, which rejects on `Content-Length` before touching the stream and then enforces the limit while reading, aborting the moment the accumulated size crosses it. Because `Content-Length` is absent on chunked bodies and can understate the real size, the streamed bound is the actual control; the header check is just an early exit. The vetted text is forwarded as-is instead of being re-serialized, dropping the second allocation. Uncapped curated proxies keep their existing path.

Covered by three tests: an oversized parsed body, an oversized `Content-Length` carrying invalid JSON (a 413 rather than a 400 proves the check precedes parsing), and an endless chunked stream with no `Content-Length` that aborts well before exhausting memory.

PR #1195, opened 2026-08-10
[Fix] Reuse compute provider clients across sleep-check ticks

mrubens:
## Problem

The sleep check runs every 60 seconds and built a brand-new compute provider client on every tick that found active or idle runs. Each client carries retained SDK state: connection channels, and the adapters' static sandbox caches (e.g. `ModalClient`'s 100-entry LRU of `Sandbox` objects) pin the client instance that created each cached entry, keeping dead per-tick clients reachable.

On deployments that always have at least one active or idle run, this leaked roughly 1.5 MB per minute of retained heap. Under the default 512 MB heap cap (#1139), the bullmq service filled its heap on a ~5.5 hour cycle and died with `FATAL ERROR: Ineffective mark-compacts near heap limit`, taking the scheduler, sleep checks, and the Discord gateway down until the platform restarted it. Mostly idle deployments take the early return before any client is built, which is why the leak only surfaced on busy ones. Before the heap cap existed, the same leak showed up as unbounded memory drift instead of crashes.

## Change

`sleep-check.ts` now keeps one client per provider in a module-level cache reused across scheduler ticks. The resolved provider env values are fingerprinted, so a credential rotation (or any provider settings change) rebuilds the client on the next tick instead of serving a stale one. Docker keeps its env-free construction.

Unchanged: the per-run memoization map, the frequency of `resolveComputeProviderEnvValues` DB reads, and the constructed client options (`envFallback` shape is identical). Only the repeated client construction stops.

`clearSleepCheckClientCache()` is exported for tests; both test suites reset it in `beforeEach` so mock clients cannot leak across cases.

## Testing

- Two new tests: the client is constructed once across two scheduler runs, and a changed env fingerprint rebuilds the client with the new `envFallback`.
- All 40 sleep-check tests and the full `@roomote/bullmq` suite (156 tests) pass; `tsc --noEmit` clean.

## Follow-ups (not in this PR)

- The controller builds a client per task spawn — same pattern at a much lower rate; worth the same treatment eventually.
- Once this ships fleet-wide, temporarily raised `NODE_OPTIONS` heap overrides can return to the 512 MB default.

### DIFF
--- a/apps/bullmq/src/scheduled-jobs/__tests__/sleep-check.test.ts
+++ b/apps/bullmq/src/scheduled-jobs/__tests__/sleep-check.test.ts
@@ -183,7 +183,12 @@ vi.mock('@roomote/db/server', () => ({
 }));
 
 // Import after mocks are set up.
-import { sleepCheckJob, sleepTaskRunNow } from '../sleep-check';
+import {
+  clearSleepCheckClientCache,
+  sleepCheckJob,
+  sleepTaskRunNow,
+} from '../sleep-check';
+import { resolveComputeProviderEnvValues } from '@roomote/db/server';
 
 /**
  * Mock the sequential DB select queries in sleepCheckJob.
@@ -224,6 +229,7 @@ function mockJobQueries({
 describe('sleepTaskRunNow', () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    clearSleepCheckClientCache();
     transactionFn.mockImplementation(async (callback) =>
       callback({ update: updateFn }),
     );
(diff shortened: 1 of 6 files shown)


mrubens:
Addressed in 5b8ad4b by making the sandbox caches per-instance in all three cached adapters (modal, daytona, e2b) rather than namespacing the static ones. The static caches only existed because clients used to be throwaway per-tick objects; with the client now long-lived, an instance cache keeps the same hit rate on the hot path while making cached handles die with the client that created them. That fixes both halves of the finding: a rebuilt client starts with a clean cache (no stale-credential handles served until TTL), and no cached entry can pin a dead client's SDK graph — which was the retention mechanism behind the original leak. The modal test's afterEach static-cache clear is removed as obsolete. All 212 compute-providers tests pass, tsc clean.

PR #1399, opened 2026-08-16
fix Brain digest SSE parsing

mrubens:
## Summary

- recognize MCP server-sent events even when an `event:` field precedes `data:`
- preserve ordinary JSON response parsing
- cover query and synthesis responses with the server-sent event shape used by gbrain

## Root cause

The nightly digest parser classified a response as SSE only when the body began with `data:`. MCP responses can begin with `event: message`, causing the valid stream to be passed to `JSON.parse` as a plain JSON document.

## Impact

Nightly Brain digests can consume gbrain MCP responses without failing before synthesis.

## Validation

- focused Brain maintenance suite: 12 passing
- BullMQ TypeScript check
- changed-file ESLint
- BullMQ formatting check
- repository pre-push checks

### DIFF
--- a/.docker/gbrain/entrypoint.sh
+++ b/.docker/gbrain/entrypoint.sh
@@ -258,13 +258,14 @@ fi
 
 # Roomote's collectors write through gbrain's MCP API. Pointing the default
 # source at a real directory makes every successful put_page also render a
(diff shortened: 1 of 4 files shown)


mrubens:
Superseded by a fresh branch based on develop to avoid carrying the prior squash-merged branch ancestry.

PR #1401, opened 2026-08-16
fix Brain digest evidence bounds

mrubens:
## What this fixes

The nightly digest first asks gbrain for pages in the eligible date window, then asks gbrain to synthesize those pages. The second step is not actually constrained to that set: gbrain v0.45 includes `since` and `until` in the synthesis prompt, but its gather phase performs a new, unbounded retrieval. That is how an older page can enter a digest even when the initial query was correct.

## How it works now

- Run one bounded gbrain query and keep the returned page excerpts, not just their slugs.
- Validate each result against its `effective_date` and exclude previously generated synthesis pages.
- Send that exact evidence set through the existing Roomote Brain inference gateway for synthesis.
- Require structured output and reject any citation whose slug was not in the supplied evidence.
- Cap evidence at 4,000 characters per page and 60,000 characters total.
- Start the gbrain query one millisecond before the first eligible UTC date. Its lower bound is strict, so this includes midnight without admitting the entire previous day.

The inference call still uses the existing Brain gateway, so provider credentials, model mapping, and deployment-specific overrides remain centralized. If synthesis or citation validation fails, no digest or digest watermark is written; the independent built-in Brain maintenance cycle still runs.

## Tests

The focused maintenance suite covers:

- bounded query timestamps
- evidence passed to the Brain inference gateway
- exclusion of out-of-window and generated pages
- rejection of citations outside the supplied evidence
- no digest or watermark on synthesis failure
- maintenance continuing when digest generation fails

Validated with 14 focused tests, BullMQ type checking, changed-file ESLint, formatting, and the repository pre-push checks.

### DIFF
--- a/apps/bullmq/src/scheduled-jobs/__tests__/brain-maintenance.test.ts
+++ b/apps/bullmq/src/scheduled-jobs/__tests__/brain-maintenance.test.ts
@@ -13,8 +13,13 @@ const {
 }));
 
 vi.mock('@roomote/sdk/server', () => ({
+  getBrainGatewayToken: () => 'brain-gateway-token',
   resolveBrainConnection: mockResolveConnection,
   resolveBrainInferenceProvider: mockResolveProvider,
+}));
+
+vi.mock('@roomote/env', () => ({
+  Env: { TRPC_URL: 'http://api.test:3001' },
 }));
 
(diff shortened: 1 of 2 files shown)


mrubens:
@roomote review

PR #1528, opened 2026-08-22
[Fix] Re-harden the Brain mirror credential on every boot

mrubens:
Why the `roomote-brain` mirror stopped updating on Aug 19.

**Root cause.** The brain repo's auto-push is gbrain's durability hardening (`gbrain sources harden`): a post-commit hook plus a repo-scoped credential helper whose file lands under `$HOME/.gbrain`. It was run by hand in an ssh session on Aug 18, where `$HOME` was `/root`, which is not on the volume. The next deploy rebuilt the container, the hook kept firing on every write-through commit, and every push failed with `could not read Username for 'https://github.com'`. From Aug 19 16:36 UTC the push log has nothing but `LOCAL-ONLY, NEEDS ATTENTION`; by today 9,698 commits were unpushed. Nothing surfaced it: the remote just looked quiet.

**Fix.** The service env still carries `GBRAIN_GITHUB_PAT`, which is exactly what the hardening reads. The entrypoint now re-runs `gbrain sources harden --all --no-cron` on every boot (when the PAT is set), under the entrypoint's volume-anchored `HOME=/data`, so the credential file is durable and a freshly provisioned brain is hardened as soon as the PAT is configured. It is idempotent by design (the dry run on the live brain reported hook, helper, and rules "already current" and only the credential missing), runs after the server is up because it needs the source registry, and is never fatal.

**Live repair.** I re-ran the hardening on the preview brain over `railway ssh` under `HOME=/data` and pushed the backlog; results in a comment below.

### DIFF
--- a/.changeset/brain-mirror-credential.md
+++ b/.changeset/brain-mirror-credential.md
@@ -1,0 +1,5 @@
+---
+"@roomote/bullmq": patch
+---
+
+Re-harden the Brain's git mirror on every gbrain container boot when GBRAIN_GITHUB_PAT is set, so the push credential survives redeploys instead of living in the container's ephemeral home directory and silently stranding every commit after the next deploy.
--- a/.docker/gbrain/entrypoint.sh
+++ b/.docker/gbrain/entrypoint.sh
@@ -379,9 +379,32 @@ echo "[gbrain-entrypoint] starting gbrain serve on :$PORT (full surface)"
 gbrain serve --http --port "$PORT" --bind "${GBRAIN_BIND:-0.0.0.0}" --surface full &
 SERVER_PID=$!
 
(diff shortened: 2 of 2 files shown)


mrubens:
Live repair on the preview brain (over `railway ssh`, `HOME=/data`):

- `gbrain sources harden default --no-cron`: credential wired to `/data/.gbrain/git-credentials` (0600) with the repo's helper repointed there, hook/helper/rules reported already current, push-probe confirmed auth.
- `git push origin HEAD:master`: `1f6546940..b56131250`, 9,698 commits; 0 unpushed afterwards. The remote was a clean ancestor, so it fast-forwarded.
- The hardening's own `pull` step logged "needs attention" because it runs before the credential is wired on a first pass; it is clean on the next run.

PR #1721, opened 2026-08-27
[Feat] Route Brain embeddings and rerank to a self-run inference upstream

mrubens:
## What changed

Lets a deployment run its own embedding and reranking models for the Brain.

**Why the gateway is the right seam:** gbrain sends embeddings, rerank, *and* chat synthesis to a single base URL (the deployment's Brain gateway), so pointing gbrain itself at an inference server would break synthesis — embedding servers have no `/v1/chat/completions`. The gateway already resolves per request, so it gains per-path upstream overrides instead:

- `R_BRAIN_EMBEDDINGS_UPSTREAM_URL` / `R_BRAIN_RERANK_UPSTREAM_URL` — any OpenAI-compatible server; each path routes independently; chat always stays with the configured provider.
- `R_BRAIN_INFERENCE_UPSTREAM_API_KEY` — one optional bearer key for both (a private-network upstream may run authless; the existing header denylist already guarantees the gateway token itself is never forwarded to any upstream).
- Model names pass through **unrewritten** on override paths: the upstream owns its model registry. This is deliberate and load-bearing — every Brain is locked to its embedding model at creation, so a served model name must mean exactly one thing forever.
- The "rerank requires OpenRouter" gate admits a configured rerank upstream.

**Self-host:** a bundled `infinity` service (CPU, `michaelf34/infinity`) behind the opt-in `local-inference` compose profile serves `/v1/embeddings` + `/v1/rerank` from one process with a model-cache volume. Enable with `COMPOSE_PROFILES=brain,local-inference` plus the two URLs (documented in `.env.production.example`).

Reranker defaults are untouched.

## How it was tested

5 new gateway tests alongside the existing 9 (all 14 pass): embeddings route to the upstream with the bearer key and *without* model rewriting or provider resolution; rerank works with a non-OpenRouter provider when the upstream is set (including trailing-slash normalization); no auth header is invented for keyless upstreams; chat still resolves the provider even with both overrides configured; caller auth still gates the local path. Env package tests pass (80); `check-types`, `format:check`, `oxlint` clean. Compose structure validated; the interpolation warnings during full `docker compose config` are pre-existing required-var prompts unrelated to the new service.

### DIFF
--- a/.env.production.example
+++ b/.env.production.example
@@ -143,11 +143,21 @@ DEFAULT_COMPUTE_PROVIDER=docker
 # Optional comma-separated GitHub App slugs that are also trusted as Roomote-managed.
 # R_GITHUB_ADDITIONAL_APP_SLUGS=
 # Self-run Brain inference (embeddings/rerank stay on your hardware; chat
-# synthesis keeps using the configured provider). With the bundled service:
-# COMPOSE_PROFILES=brain,local-inference and the URLs below.
+# synthesis keeps using the configured provider). With the bundled service,
+# set COMPOSE_PROFILES=brain,local-inference and ALL of the settings below —
+# the model names and dimensions must match what the inference server
+# serves, and the embedding pair is create-time: set everything BEFORE the
+# Brain's first boot. gbrain's defaults (text-embedding-3-small, 1536) name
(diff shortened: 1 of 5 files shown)


mrubens:
@roomote review

PR #1726, opened 2026-08-27
[Fix] Discover Notion pages the search index never surfaces

mrubens:
## What changed

The Notion collector discovered pages exclusively via `POST /v1/search`. Per [Notion's own documentation](https://developers.notion.com/reference/search-optimizations-and-limitations), search is *"not guaranteed to return everything"* and is explicitly not for *"exhaustively enumerating through all the documents that a bot has access to"* — only **directly-shared** content is guaranteed to be returned; pages that inherit access through a shared parent may never appear. Observed in production: a page fully accessible to an integration by direct fetch, yet absent from search results and therefore never ingested, with the collector otherwise healthy and its backfill complete.

The scan state machine gains a **traversal phase** after each sweep + reconcile cycle (`sweep → reconcile → traverse → idle`):

- Seeds from the existing inventory, iterated by a durable `brain_collector_items` id cursor (new `listBrainCollectorItemsAfter` helper) — no API cost and no queue growth for seeding.
- For each page: walks `blocks/{id}/children` (paginated), ingesting `child_page` blocks the inventory doesn't track, descending into container blocks (toggles, columns, …) to a bounded depth.
- For `child_database` blocks: resolves the database's data sources and enumerates rows via `data_sources/{id}/query` — the enumeration path Notion recommends instead of search.
- Newly discovered pages are fetched, emitted through the normal page pipeline, inventoried with the scan timestamp (so the next reconcile doesn't tombstone them), and their own subtrees are queued.
- Budgets: 12 requests per pass, bounded pending list (300, drops logged), cursor persisted in `brain_sync_state` every pass — a traversal cycle spans as many ticks as it needs and resumes exactly where it stopped. Rate-limit errors bubble through the existing collector backpressure contract.

The incremental search-based sweep is unchanged — it remains the freshness fast path; traversal closes the completeness hole.

## How it was tested

5 new tests in `brain-notion.test.ts`: a child page absent from search is discovered and ingested; an inventory-tracked child is not re-fetched (dedup); a pass that exhausts its request budget persists mode `traverse` with pending work and the advanced seed cursor; child databases route through data-source queries; an exhausted inventory completes back to `idle`. The reconcile-completion test now pins the new `reconcile → traverse` handoff. Full bullmq suite: 406 passed. `check-types` (bullmq + db), `format:check`, `oxlint` clean.

### DIFF
--- a/apps/bullmq/src/scheduled-jobs/__tests__/brain-notion.test.ts
+++ b/apps/bullmq/src/scheduled-jobs/__tests__/brain-notion.test.ts
@@ -1,6 +1,10 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-import { listBrainCollectorItemsBefore } from '@roomote/db/server';
+import {
+  listBrainCollectorItemsAfter,
+  listBrainCollectorItemsBefore,
+  listBrainCollectorItemsBySlugPrefix,
+} from '@roomote/db/server';
 import { NotionApiError } from '@roomote/sdk/server/notion-api';
 
 import {
@@ -10,6 +14,7 @@ import {
   buildNotionUserReferences,
   buildUnavailableNotionPage,
   collectNotionReconciliation,
+  collectNotionTraversal,
   type NotionSearchPage,
   type NotionUserIdentity,
 } from '../brain-collectors/notion-pages';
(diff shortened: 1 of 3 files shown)


mrubens:
@roomote review

PR #2034, opened 2026-09-02
[Fix] Let OpenCode install its own zod into the shared Fast tools directory

mrubens:
## Problem

On the nightly, Fast replies that carry array arguments (`imageArtifactIds`, `suggestions`) are being rejected by OpenCode's argument validation; the model reports "the reply tool is rejecting the array parameters" and falls back to prose.

Cause: #2026 seeded a `package-lock.json` in the shared tools directory so OpenCode would skip its per-directory dependency install, and symlinked the app's `zod` (3.25) for the tool files. Before #2026, that install ran on every conversation directory and replaced the symlink with zod 4.1.8 (the version `@opencode-ai/plugin` depends on). OpenCode validates tool arguments with zod 4, and zod 3 schema objects don't survive that for arrays.

## Change

Drop the seeded lockfile and the zod symlink. OpenCode installs `@opencode-ai/plugin` and its zod once per shared directory per host (the directory is content-addressed, so once per release), and every later conversation reuses it. The per-conversation saving from the shared directory is unchanged; the one-time install now lands on the first Fast turn after a deploy (~1–5 s, network to the npm registry, same as every conversation paid before #2026).

## Verification

- Live local turn asking for `send_chat_reply` with `imageArtifactIds: []`: the reply posted, OpenCode installed zod 4.1.8 into the new shared directory on first boot (3.7 s), and the turn's diagnostics are normal.
- `src/server/fast-agent` suite passes; typecheck and lint clean.

### DIFF
--- a/packages/cloud-agents/src/server/fast-agent/fast-agent-native-tool-bridge.ts
+++ b/packages/cloud-agents/src/server/fast-agent/fast-agent-native-tool-bridge.ts
@@ -5,19 +5,14 @@ import {
 } from 'node:http';
 import {
   chmodSync,
-  lstatSync,
-  readdirSync,
+  existsSync,
   mkdirSync,
   rmSync,
-  statSync,
-  symlinkSync,
-  unlinkSync,
   writeFileSync,
 } from 'node:fs';
 import { tmpdir } from 'node:os';
-import { dirname, isAbsolute, join } from 'node:path';
+import { join } from 'node:path';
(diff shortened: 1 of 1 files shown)


mrubens:
Verified end to end on a local turn: the model called send_chat_reply with imageArtifactIds: ["art_123"], OpenCode recorded the tool part as completed with that input, and the reply reached the adapter with the array intact. The shared directory now contains zod 4.1.8 installed by OpenCode.

PR #2139, opened 2026-09-03
[Fix] Fast turns interrupted after starting a task no longer ask the user to resend

daniel-lxs:
## What changed

- A Fast turn interrupted after it had already started a task no longer closes out with "Please send it again." It says the task it started is still running and will report back in the thread.
- The closeout's marker records which task(s) the turn started (`delegatedTaskIds`), and the unresolved-request carry-over skips such turns, so the next message is treated as a new request rather than a redo of the delegated one.
- Applies to every interruption closeout variant (restart, retry-notice replacement) and is decided by launches made in the interrupted turn itself, not by tasks that were already active in the session.

## Why

Observed on staging: a turn started a task, a deploy killed the turn twenty seconds later, and the closeout told the user to send the request again. Following that instruction would have started a second task. The one that existed kept running and its result was always going to arrive in the thread through the parent-event queue; the parent turn's only remaining job was to say so.

The next message did resume the request through the carry-over and, fortunately, the model found the running task instead of relaunching. This change removes the dependency on that judgment: the closeout is accurate on its own and the carry-over no longer offers the request back.

## What it does not change

- Turns that acted without starting a task (a failed launch, a memory write, a task message) still get the honest "send it again" closeout, since nothing is running on the user's behalf.
- Turns that had not acted are still handed back to the queue and resumed silently, as before.
- No schema change. The marker is a metadata field on the existing closeout row.

## Verification

- `fast-agent-service` tests: a shutdown after a successful launch posts the delegated wording and records the task id on the closeout; a shutdown after a failed launch keeps the resend wording and records no task.
- `fast-agent-conversation-repository` real-DB test: a delegated closeout is not surfaced as an unresolved request.
- Full fast-agent suite, type check, lint, and knip pass.

### DIFF
--- a/.changeset/fast-delegated-interruption.md
+++ b/.changeset/fast-delegated-interruption.md
@@ -1,0 +1,5 @@
+---
+'@roomote/web': patch
+---
+
+Fixed Fast Sessions asking the user to resend a request after a restart interrupted a turn that had already started a task. Resending would have started a second task. The closeout now says the started task is still running and will report back in the thread, and the next message no longer treats that request as unfinished.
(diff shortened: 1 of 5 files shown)


daniel-lxs:
Superseded by https://github.com/RooCodeInc/Roomote/pull/2145, which makes every interrupted Fast turn resume automatically with a record of what its earlier attempt already did. Closing; can be reopened if that approach does not land.

PR #2719, opened 2026-09-15
[Feat] Allow service-specific credential headers on Session secret grants

daniel-lxs:
## Why this PR exists

Session secret grants could only name `authorization`, `x-api-key`, or `api-key` as the header that carries the key. Many services authenticate with their own header: GitLab's `PRIVATE-TOKEN`, Shopify's `X-Shopify-Access-Token`, Supabase's `apikey`, Vault's `X-Vault-Token`. Those services could not be approved at all.

Stacked on #2718, which adds the API-side proxy this change extends.

## What changed

- `packages/types`: a shared credential header schema. Any RFC 7230 header token is accepted and stored lowercase, up to 64 characters. Headers that shape or route the request are refused: cookies, `host`, content framing, caching and conditional headers, `user-agent`, and anything under `proxy-`, `x-forwarded-`, or `sec-`. A scheme (`Bearer`, `Basic`, `Token`) is still only accepted on `authorization`, which the SDK already enforced.
- `packages/db`: the two `header_name` columns lose their three-value TypeScript type. They were plain `text` already, so there is no migration.
- HTTP integrations broker: validates the grant's header with the shared predicate instead of the hardcoded list, and injects into whatever header the grant names.
- API proxy route: looks for the substitute in every request header rather than three fixed slots, strips whichever header carried it plus the grant's own header, and injects the real key into the grant's header. Ambiguity rules are unchanged: exactly one substitute value.
- Fast `prepare_session_secret` tool schema: `headerName` becomes a described string; the server-side prepare schema still validates it, so a bad name fails exactly as before.
- Docs: the Session secrets page gets a "Header names" section. The Iron gateway README notes that the connector path still accepts only the three standard slots; grants on other headers are served by the API proxy. The gateway's Go code is deliberately untouched, since that path is slated for removal.

## How it was tested

- Types: new cases for accepted names (including mixed case and whitespace normalization) and refused ones (`cookie`, `host`, `content-type`, `proxy-authorization`, `x-forwarded-for`, `sec-fetch-mode`, empty, spaces, over-long, pseudo-headers).
- API real-database suites: broker forwards `private-token: <real key>` for a custom grant and refuses a scheme on it or a `cookie` grant; the proxy serves a `PRIVATE-TOKEN` grant whether the client presented the substitute in `private-token`, `authorization`, or an unrelated header, and never forwards the presenting header. All 258 API session-egress and integrations tests pass.
- Fast agent service and tool policy suites (344 tests), the web secrets route and form suites, and typecheck for types, db, sdk, api, cloud-agents, web, worker, and controller all pass. oxlint and knip clean.


daniel-lxs:
Folded into #2718 as commit 6db586819; one review instead of a stacked pair.

PR #2726, opened 2026-09-15
[Fix] Enable the owner experiment on the SDK session-egress test fixture

daniel-lxs:
## Why this PR exists

`develop` CI is red after #2720. That PR gates substitute minting on the Session owner's Session secret tools setting, and the SDK's session-egress integration fixture creates owners without it, so fourteen cases fail with `run_not_eligible`. The API fixtures were updated in #2720; the SDK fixture was missed because those suites did not run in the PR's local verification.

## What changed

- The SDK integration test's owner is created with `session_secret_tools_enabled: true`, matching the API fixtures. One line plus a comment; no product code.

## How it was tested

- `packages/sdk` session-egress integration and unit suites: 40 tests pass against the real test database. Typecheck clean.


daniel-lxs:
Superseded: the same fixture change landed on develop with #2723, so this is no longer needed.

PR #2817, opened 2026-09-16
[Improve] Deference is about the user's choices, not a reason to ask before looking

mrubens:
## Problem

The deference paragraph added in #2754 tells the agent to offer rather than announce when it wants to deviate from something the user chose. Read against a terse trigger, the model takes it as "ask before assuming". A one-line automated Slack mention of the form `*<suite> — tests need updates*` used to make the agent pull the surrounding thread, look up recent failures, and launch a task. With the paragraph in place the same opening prompt now gets "Which tests need updating? Please share the ticket, PR, or failing test names" and the turn stops without looking anything up. The ingestion code for that path is unchanged; this paragraph is the only prompt change that plausibly explains the flip.

## Change

The paragraph now scopes the deference to *a method or decision the user chose* and adds the counterweight: when a request names work without specifics (a bare title, a one-line trigger, an automated notice), find the specifics first from the conversation, the repositories, recent failures, and memory, and ask only when that search leaves the work genuinely ambiguous. The tone rules from #2754 are unchanged.

Prompt-only; the shared style guidance feeds Fast and the standard task prompts alike.

## Validation

`fast-agent-prompt.test.ts` asserts the new sentence alongside the existing deference assertions; 111 tests pass. `tsc`, oxfmt, oxlint clean.


mrubens:
Rewrote the description to drop the tenant identity, the workflow title, and the session history; the rationale is kept in general terms.

mrubens:
@roomote-community the description has been rewritten to remove the identity and history; please re-review.

PR #2820, opened 2026-09-16
Deliver an automated author's thread replies to its Session

mrubens:
## Problem

A workflow or app that summons Roomote with a one-line mention usually posts its real content as thread replies a second or two later. The mention starts the Fast turn immediately, so the thread fetch at turn start sees only the headline. The replies themselves never reach the Session: they are bot-authored (`bot_id`/`app_id`, often no linked Slack user mapping), so the unmentioned-reply path drops them at the user-mapping step and nothing notifies the Session afterwards. The agent ends up asking what it should be looking at while the answer sits two messages down the thread.

Companion to #2818, which fixes what gets extracted from those messages. This PR fixes whether the Session sees them at all.

## Change

`apps/api/src/handlers/slack/events/message-entry.ts`

- New `maybeRouteAutomatedThreadReply`, run before the human routing paths. A thread reply is delivered to Fast under the automation launch identity (the same identity the automated mention used, with the automation initiator for delegated tasks) when all of these hold:
  - the reply is bot- or app-authored and not Roomote's own;
  - it does not mention Roomote (that is an automated mention already) or anybody else;
  - it lands within `AUTOMATED_THREAD_REPLY_WINDOW_SECONDS` (15 min) of the root;
  - the thread is bound to a Fast Session;
  - the root message was authored by the same `bot_id` or `app_id`.
- Delivery goes through `startFastAgentResponse`, so the existing admission logic applies: an active turn is steered with the new content, an idle Session gets a new turn. A burst of replies collapses into steers of the first one.

Everything else (other bots, late replies, replies addressed to a person, threads without a Session) falls through to the existing behavior unchanged.

## Validation

New `message-entry-automated-thread-reply.test.ts`: delivers a same-app reply with its block context under the automation identity; ignores no-Session threads, other apps, late replies, Roomote's own replies, and replies that address someone else. All Slack handler tests pass (34 files, 269 tests); `tsc`, oxfmt, oxlint clean.


mrubens:
Closing without merging. #2818 gives the first turn the full original message (the attachment carries the task, repository, and PR link), which is enough to act on. Delivering an automated author's later thread replies as follow-up turns also fires on late status updates from other bots, so the broader mechanism is not worth its side effects for now. Branch kept in case the need returns.

PR #3029, opened 2026-09-20
fix(worker): allow slower OpenCode startup

parland:
## What changed

- Increase the bounded OpenCode readiness allowance from 30 seconds to 90 seconds.
- Name the startup timeout so it is distinct from request, health, and stall timeouts.

## Why

Cold repository-attached startup can legitimately approach or exceed 30 seconds while the server remains healthy. At the current boundary, small host or workspace-preparation variance causes the worker to terminate the subprocess just before it becomes ready. A 90-second bound preserves fail-closed startup while removing that race.

## Validation

- `readiness.test.ts`: 4 passed
- `@roomote/worker` TypeScript check: passed
- `@roomote/worker` lint: passed
- repository pre-push oxlint, residual lint, fast type checks, and knip: passed

Local validation used Node 23 because Node 22/24 was unavailable on the operator host; the repository emitted its expected engine warning. CI remains authoritative for the supported runtime.


mrubens:
@parland thanks! I’ll merge as soon as you accept the CLA

parland:

parland:
recheck

